PR #7, opened 2022-04-27
Normalize RST labels to remove duplicates (for one option)

felixfontein:
docutils normalizes labels to targets by making them lowercase and normalizing whitespace. We do this for all labels for an option to avoid collisions. (You can still have collisions if you have two distinct options that differ by case, for example.)


briantist (approved):
(as discussed in IRC) I would just add the reasoning for the urlencode change to changelog, otherwise this is great!

Context for other readers: https://docs.ansible.com/ansible/latest/collections/amazon/aws/aws_s3_module.html#parameter-s3_url

This module has an option with an alias that differs only in case, while RST treats labels as case insensitive, so it results in a duplicate which cases a warning.

felixfontein:

PR #8, opened 2022-05-05
Support plugin seealso

felixfontein:
Defined in semantic markup spec: https://hackmd.io/VjN60QSoRSSeRfvGmOH1lQ?both

This is needed for https://github.com/ansible/ansible/pull/77687. It can mainly be used by plugins that are not validated by ansible-test's validate-modules, i.e. test and filter plugins.


samccann (commented):
Can't comment on the code but +1 the idea.

felixfontein:
@briantist @acozine @samccann thanks a lot for reviewing this and taking a look at this!

PR #10, opened 2022-06-18
Avoid repeating input argument when it is mentioned as positional

felixfontein:
Basically fixes https://github.com/ansible/ansible/pull/77725 on our side. (I still think this is wrong and should be highlighted by a linter, but there is no linter and if core doesn't change it, we have to accept it as-is...)


briantist (approved):
I also think `_input` being treated as positional is wrong (commented in detail on the other PR), but if it doesn't go our way, I approve of this implementation of that decision (we should wait until that one is decided I guess?)

felixfontein:
It's also worth having this merged as there's no linting tool that prevents collection maintainers from doing the same. (They'll likely then complain that we render the docs wrongly...)

felixfontein:

PR #12, opened 2022-07-04
Extended plugin docs lint

felixfontein:
Add an option `--plugin-docs` to the `lint-collection-docs` subcommand that creates RST files out of the plugin docs and lints them. It also prints all rendering errors it finds as warnings.


briantist (approved):

felixfontein:

PR #14, opened 2022-07-24
Follow REUSE specification

felixfontein:
https://reuse.software/spec/


felixfontein:
CC @gotmax23

gotmax23:
I have been summoned! I will try to look at this later.

src/antsibull_docs/data/docsite/list_of_collections.rst.j2
```diff
@@ -1,3 +1,9 @@
+{#
+  Copyright (c) Ansible Project
+  GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+  SPDX-License-Identifier: GPL-3.0-or-later
+#}
+
```
gotmax23, line 6: This (and some of the other jinja2 files) use Jinja2 comments that don't show up in the rendered file. Is that on purpose?

src/sphinx_antsibull_ext/antsibull-minimal.css.license
```diff
@@ -0,0 +1,3 @@
+GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+SPDX-FileCopyrightText: Ansible Project
+SPDX-License-Identifier: GPL-3.0-or-later
```
gotmax23, line 3: Is this supposed to be included both here and in the file?

src/sphinx_antsibull_ext/css/antsibull-minimal.scss
```diff
@@ -1,4 +1,6 @@
 /* Copyright (c) Ansible and contributors */
+/* GNU General Public License v3.0+ (see https://www.gnu.org/licenses/gpl-3.0.txt) */
+// SPDX-License-Identifier: GPL-3.0-or-later
```
gotmax23, line 3: Different comment styles

stubs/ansible/__init__.pyi
```diff
@@ -0,0 +1,3 @@
+# Copyright (c) Ansible Project
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
```
gotmax23, line 3: It looks like `reuse lint` doesn't care about blank files (i.e. they don't need identifiers). I didn't check this before, hence I'm only pointing it out here :).

tests/functional/schema/good_data/one_lookup_results.json.license
```diff
@@ -0,0 +1,2 @@
+SPDX-FileCopyrightText: Ansible Project
+SPDX-License-Identifier: GPL-3.0-or-later
```
gotmax23, line 2: This and some of the other `.license` files don't have both types of license identifier comments.

src/antsibull_docs/data/docsite/list_of_collections.rst.j2
felixfontein: Yes, it's intentional. The generated RST files mainly contain information from the collection the docs are generated of.

src/sphinx_antsibull_ext/css/antsibull-minimal.scss
felixfontein: `/* */` comments are included in the generated CSS files, `//` comments are not. This is important here since `SPDX-License-Identifier` would end up in a way that REUSE cannot use (because there's something after the license name in the same line), but we do want the copyright and readable license information in the generated file.

src/sphinx_antsibull_ext/antsibull-minimal.css.license
felixfontein: Kind of. `SPDX-License-Identifier` is not a duplicate; removing `SPDX-FileCopyrightText` results in linting failures (the copyright text is no longer found). The only thing that's something we can remove here is the first line, but for consistency I would keep it.

stubs/ansible/__init__.pyi
felixfontein: I also noticed that, but I guess it doesn't hurt either in case someone ever wants to add code in here.

tests/functional/schema/good_data/one_lookup_results.json.license
felixfontein: Thanks, fixed in 52a67d7a92309bdd7358d940171265b5fea33310.

src/antsibull_docs/data/docsite/list_of_collections.rst.j2
gotmax23: Wouldn't the generated RST files still be considered GPL, as the template is?

felixfontein: I guess that's a question that will spark very interesting legal discussions among legal experts ;-) I have no idea. I can think of reasons for both "yes" and "no", so here I would avoid trying to take a side on this question by avoiding to insert copyright or license statements into the generated RST.

gotmax23: > I would avoid trying to take a side on this question by avoiding to insert copyright or license statements into the generated RST.

I guess. We might want to re-license the files under a very permissive license such as [MIT-0](https://spdx.org/licenses/MIT-0.html) (a MIT license variant that doesn't require any attribution), but that's another conversation.

You can resolve this thread.

src/sphinx_antsibull_ext/antsibull-minimal.css.license
gotmax23: Resolve

src/sphinx_antsibull_ext/css/antsibull-minimal.scss
gotmax23: Resolve

src/antsibull_docs/data/docsite/list_of_collections.rst.j2
felixfontein: That also might require quite some effort, since these templates are based on templates in ansible/ansible, which have been edited by quite a few of folks over some years. Figuring out who was involved and asking all of them will be some effort. Doing that would be a good idea though IMO, but I don't want to do that anytime soon :)

gotmax23 (approved):
I looked over this again, and it looks good to me. I was confused about why those threads were resolved when I didn't do it, but then I realized you got to it this quickly :).

PR #18, opened 2022-08-03
Fix JSON string escaping

felixfontein:
As reported by @LvffY on #ansible-docs, in default and return value examples (which are JSONified), `<` and `>` are escaped as `\u003c` resp. `\u003e`. This is something Jinja2 does to avoid folks to shoot themselves in the foot in HTML templates. Since we properly escape the `tojson` output, we don't need this extra safety. This PR removes that extra safety.


Andersson007 (approved):

felixfontein:
@Andersson007 thanks for reviewing this!

PR #19, opened 2022-08-04
Convert 'default', 'choices', 'sample' to type; change booleans to 'true' and 'false'

felixfontein:
Implements https://github.com/ansible-community/community-topics/issues/116

### DIFF
--- a/LICENSES/BSD-2-Clause.txt
+++ b/LICENSES/BSD-2-Clause.txt
@@ -1,0 +1,8 @@
+Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.
+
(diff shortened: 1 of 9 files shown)


felixfontein:
The part on converting `default`, `sample`, and `choices` is necessary since there are quite a few modules and plugins which use strings like `"yes"` to provide default values (or choices, or samples), instead of proper boolean values. Without this change, there were quite a few cases where for a boolean parameter / return value the choices are `true` and `false`, but the default is `yes`, `False`, or something else. (Strings with `False`, `True`, `false`, or `true` in them are already showing up wrong at the moment.)

felixfontein:
You can find generated docs at https://ansible.fontein.de/collections/

Andersson007 (approved):
@felixfontein thanks for implementing the decision!

felixfontein:
@Andersson007 thanks for reviewing this!

PR #25, opened 2022-08-16
Show 'might already be installed as part of ansible' disclaimer only on official docsite, but also for roles

felixfontein:
Result for a non-official docsite: https://ansible.fontein.de/collections/community/general/alternatives_module.html#ansible-collections-community-general-alternatives-module no longer says this.


samccann (approved):
lgtm

felixfontein:
@samccann thanks a lot for reviewing this one as well :)

PR #29, opened 2022-08-22
Use collection's issue tracker URL instead of collection URL when suggesting to file a bug with the collection

felixfontein:
As discussed on IRC.


briantist (approved):

felixfontein:

PR #32, opened 2022-09-03
Make docs processing more robust

felixfontein:
Currently the devel docs build fails with https://pastebin.com/K1u95BXx.

### DIFF
--- a/src/antsibull_docs/cli/doc_commands/stable.py
+++ b/src/antsibull_docs/cli/doc_commands/stable.py
@@ -155,8 +155,7 @@ def normalize_plugin_info(plugin_type: str,
     for field in ('doc', 'examples', 'return'):
         try:
             schema = DOCS_SCHEMAS[plugin_type][field]  # type: ignore[index]
-            default_value: t.Any = None if field == 'examples' else {}
-            field_model = schema.parse_obj({field: plugin_info.get(field) or default_value})
+            field_model = schema.parse_obj({field: plugin_info.get(field)})
         except ValidationError as e:
             if field == 'doc':
                 # We can't recover if there's not a doc field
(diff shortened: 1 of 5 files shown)


src/antsibull_docs/cli/doc_commands/stable.py
```diff
@@ -166,7 +169,8 @@ def normalize_plugin_info(plugin_type: str,
             # But we can use the default value (some variant of "empty") for everything else
             # Note: We looped through doc first and returned an exception if doc did not normalize
             # so we're able to use it in the error message here.
-            errors.append(f'Unable to normalize {new_info["doc"]["name"]}: {field}'
+            plugin_name = (new_info["doc"] or {}).get("name") or "???"
```
briantist, line 172: I would appreciate a comment in the code explaining how/why the string `"???"` could end up here

briantist (approved):
Just a non-blocking request for a comment on the confusing  `???` value

src/antsibull_docs/cli/doc_commands/stable.py
felixfontein: I added a comment in 3b0a9328290d1138d386d09feb3ae78601d7beee, and at the same time decided to rewrite that function a bit.

briantist (approved):

felixfontein:
Hmm, something isn't right yet. Need to look at this some more.

src/antsibull_docs/cli/doc_commands/stable.py
felixfontein: Actually I undid most changes to this function.

felixfontein:
I'm now happy with the actual docsite output.

(This project definitely needs better tests - probably we need to include a test collection and the expected RST output for it.)

felixfontein:

PR #33, opened 2022-09-14
Remove duplicate tests and filters

felixfontein:
Fixes the ansible-core sanity test errors in https://github.com/ansible/ansible/pull/78674#issuecomment-1245998705.

### DIFF
--- a/changelogs/fragments/33-remove-test-filter-aliases.yml
+++ b/changelogs/fragments/33-remove-test-filter-aliases.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "Detect filter and test plugin aliases and avoid them being emitted multiple times. Instead insert redirects so that stub pages will be created (https://github.com/ansible-community/antsibull-docs/pull/33)."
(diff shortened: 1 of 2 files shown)


briantist (approved):

briantist:
This looks good to me, but what are aliases exactly? Are they just when filters/tests of different names use the same implementation? In the code comment, it looks for identical docs, which makes me wonder, how do we determine which is the "canonical" page?

felixfontein:
> This looks good to me, but what are aliases exactly? Are they just when filters/tests of different names use the same implementation?

Basically yes. (In ansible-core, symlinks are used to implement this for the documentation, and in the code the same function is exposed with different names.)

> In the code comment, it looks for identical docs, which makes me wonder, how do we determine which is the "canonical" page?

Assuming that symlinks are used for the `.yml` files, the canonical version is the one whose `name:` coincides with the plugin's name. For example, in ansible-core's repo, `lib/ansible/plugins/test/abs.yml` has `name: abs`, and `lib/ansible/plugins/test/is_abs.yml` (symlink) also has `name: abs`.

PR #36, opened 2022-09-21
Allow to specify choices as dictionary instead of list

felixfontein:
Needed for the `combine` filter in ansible-core (https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/filter/combine.yml#L26-L32).

### DIFF
--- a/changelogs/fragments/36-choices-dict.yml
+++ b/changelogs/fragments/36-choices-dict.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "Allow to specify choices as dictionary instead of list (https://github.com/ansible-community/antsibull-docs/pull/36)."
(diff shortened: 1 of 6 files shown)


felixfontein:
Result: compare https://docs.ansible.com/ansible/devel/collections/ansible/builtin/combine_filter.html vs. https://ansible.fontein.de/collections/ansible/builtin/combine_filter.html#parameter-list_merge

Also, a side effect of this change: https://docs.ansible.com/ansible/devel/collections/community/crypto/openssl_csr_module.html#parameter-version vs https://ansible.fontein.de/collections/community/crypto/openssl_csr_module.html#parameter-version

briantist:
This looks great, I saw the discussion in IRC and also never knew this was possible. Documenting choices that way is nice.

Only suggestion I have for the output is to have the choice values use `C()` format. That would be a departure from non-described choices unless we applied it to all of them... which I personally think we should, but I suppose that might need some discussion/voting.

felixfontein:
@briantist I think that should apply to `default` as well. I'll do that in a different PR though (since I think that's separate enough), once this one is merged (to avoid conflicts).

felixfontein:
@briantist @bcoca thanks for your comments!

Issue #39, opened 2022-09-24
Docs builds are failing with Sphinx 5.2.0.post0


briantist:
Sphinx 5.2.0 was just released today:
- https://github.com/sphinx-doc/sphinx/issues/10853

I haven't been able to figure out why yet, but at least in `community.hashi_vault` the docs build is failing with:

> Reason: ValueError('too many values to unpack (expected 3)')

I first noticed it after making an innocuous change to a previously passing build (adding license header comments):
https://github.com/ansible-collections/community.hashi_vault/actions/runs/3119998619/jobs/5060610693#step:8:126

In CI it's coming up on my inclusion of changelog in the docsite:

>  Theme error:
>   An error happened in rendering the page collections/community/hashi_vault/docsite/CHANGELOG.
>   Reason: ValueError('too many values to unpack (expected 3)')

When I tried to reproduce it locally, I couldn't, until I upgraded the requirements (after having installed them fresh a few hours earlier) and saw that it upgraded sphinx from 5.1.1 to 5.2.0.post0, after thatI could repro, but it came up on a different file, so I don't _think_ it's related to anything I'm doing:

> Theme error:
> An error happened in rendering the page collections/community/hashi_vault/docsite/about_hashi_vault_lookup.
> Reason: ValueError('too many values to unpack (expected 3)')

There's at least one other report of a failing build with it:
- https://github.com/sphinx-gallery/sphinx-gallery/issues/1017

AA-Turner:
Hi,

Are you able to provide a consistently failing minimal example? If this is a regression in Sphinx 5.2 I'd like to fix it sooner rather than later.

A

felixfontein:
From [this comment](https://github.com/sphinx-doc/sphinx/issues/10853#issuecomment-1257153452) it seems that this problem is not related to antsibull-docs, but to the sphinx-rtd-theme that [sphinx_ansible_theme](https://github.com/ansible-community/sphinx_ansible_theme) is based on.

I verified this myself, with the following simple patch to sphinx-rtd-theme, both my docsite (using sphinx-rtd-theme directly) and another test docs build (using sphinx_ansible_theme)
```.diff
diff --git a/sphinx_rtd_theme/layout.html b/sphinx_rtd_theme/layout.html
index 1d70119..c9c07b9 100644
--- a/sphinx_rtd_theme/layout.html
+++ b/sphinx_rtd_theme/layout.html
@@ -10,7 +10,7 @@
[...]

PR #40, opened 2022-09-25
Avoid Sphinx 5.2.0.post0 for docsite builds to work around a sphinx-rtd-theme bug

felixfontein:
This is mainly a workaround, see #39 for details.

The proper fix would be getting https://github.com/readthedocs/sphinx_rtd_theme/pull/1344 merged and into a new release ASAP.

### DIFF
--- a/src/antsibull_docs/data/sphinx_init/requirements_txt.j2
+++ b/src/antsibull_docs/data/sphinx_init/requirements_txt.j2
@@ -4,5 +4,5 @@
 
 antsibull-docs >= 1.0.0, < 2.0.0
 ansible-pygments
-sphinx
+sphinx != 5.2.0.post0  # The != 5.2.0.post0 restriction is thanks due to a bug in the sphinx-rtd-theme. Once this is fixed, this restriction can be removed.
 @{ sphinx_theme_package }@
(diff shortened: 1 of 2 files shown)


src/antsibull_docs/data/sphinx_init/requirements_txt.j2
```diff
@@ -4,5 +4,5 @@
 
 antsibull-docs >= 1.0.0, < 2.0.0
 ansible-pygments
-sphinx
+sphinx != 5.2.0  # The != 5.2.0 restriction is thanks due to a bug in the sphinx-rtd-theme. Once this is fixed, this restriction can be removed.
```
briantist, line 7: If the issue is not actually in `sphinx` itself, should we do it like this? presumably a new release of `sphinx` would not fix the issue and we'd be right back where we started.

```suggestion
sphinx < 5.2.0  # The < 5.2.0 restriction is thanks due to a bug in the sphinx-rtd-theme. Once this is fixed, this restriction can be removed.
```

felixfontein: Actually the problem isn't 5.2.0, but 5.2.0.post0. I've changed this line to avoid `5.2.0.post0`, which at least fixes the issue in CI for this repo.

briantist: ah yup, I just found that myself, was about to say the same. Perfect, thanks!

PR #34, opened 2022-09-14
Fix/improve version_added

felixfontein:
Fix escaping and collection names in roles options, and make it say `added in version 2.11 of ansible-core` instead of `added in version 2.11 of ansible.builtin`.

Examples:

- https://ansible.fontein.de/collections/ansible/builtin/file_module.html#parameter-access_time
- https://ansible.fontein.de/collections/community/crypto/openssl_csr_module.html#parameter-attributes
- https://ansible.fontein.de/collections/felixfontein/acme/acme_certificate_role.html#parameter-main--acme_certificate_renewal_on_remaining_days

PR #41, opened 2022-09-25
Fix module/role level version_added rendering

felixfontein:
There `_` is escaped and the escaping is visible in the rendered HTML, see for example https://docs.ansible.com/ansible/devel/collections/community/hashi_vault/vault_kv2_get_module.html, https://docs.ansible.com/ansible/devel/collections/community/sap_libs/sap_snote_module.html, and https://docs.ansible.com/ansible/devel/collections/cloudscale_ch/cloud/network_module.html.

Thanks to @briantist for noticing and reporting this on IRC.

### DIFF
--- a/changelogs/fragments/41-version_added.yml
+++ b/changelogs/fragments/41-version_added.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - "Avoid collection names with ``_`` in them appear wrongly escaped in the HTML output (https://github.com/ansible-community/antsibull-docs/pull/41)."
--- a/src/antsibull_docs/data/docsite/macros/version_added.rst.j2
+++ b/src/antsibull_docs/data/docsite/macros/version_added.rst.j2
@@ -18,24 +18,24 @@ ansible-core
 
 {% macro version_added_rst(version_added, version_added_collection=None, do_not_escape_first_word=False) -%}
 {%-   if version_added_collection == 'ansible.builtin' -%}
-@{ ansible_name(version_added) }@ @{ version_added | rst_escape }@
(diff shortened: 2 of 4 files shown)


felixfontein:
You can see the improved output here: https://ansible.fontein.de/collections/community/hashi_vault/vault_kv2_get_module.html#ansible-collections-community-hashi-vault-vault-kv2-get-module

briantist (approved):
thank you!

felixfontein:

samccann:
fwiw, it used to say;
 `New in version 2.5.0: of community.hashi_vault
`
And now it says:
`New in version community.hashi_vault: 2.5.0`

I prefer the older way of putting the version first. Do we know why it changed?

felixfontein:
@samccann this was changed in #34, not in this PR. The main reason this change happened was because all *other* version added texts were bad. That this one changed in this way wasn't intentional.

felixfontein:
In any case, `New in version 2.5.0: of community.hashi_vault` isn't great either, since this basically says (using Sphinx's semantics) that the feature "of community.hashi_vault" was added in version 2.5.0.

samccann:
ok thanks for the explanation. What do you think about `New in community.foo: version 2.4.0` as an enhancement in the future?

felixfontein:
@samccann what do you think about `New in community.hashi_vault 2.5.0`, as in https://ansible.fontein.de/collections/community/hashi_vault/vault_kv2_get_module.html or https://ansible.fontein.de/collections/ansible/builtin/dnf_module.html ?

felixfontein:
(I created #43 for this.)

PR #47, opened 2022-09-30
Fix `lint-collection-docs --plugin-docs` subcommand

felixfontein:
Right now it seems to be broken, and it wasn't even used in the tests :(


felixfontein:
@briantist thanks a lot for reviewing this!

PR #54, opened 2022-10-06
Improve CSS to avoid left column to be too wide

felixfontein:
Fixes parts of https://github.com/ansible-collections/amazon.aws/issues/1101.


felixfontein:
You can find this version deployed to https://ansible.fontein.de/collections/. I didn't find any module on my docsite with such a list of aliases as that module in amazon.aws though...

PR #70, opened 2022-11-24
remove compatibility code for old Pythons

gotmax23:


felixfontein (changes requested):
This is a breaking change and can only be merged for 2.0.0. Marking with a negative review so we don't accidentally merge this.

felixfontein:
@gotmax23 do you want to rebase to remove all parts that are no longer needed, or should we create a new PR with similar changes?

felixfontein (approved):

felixfontein:
@gotmax23 thanks for doing this and rebasing :)

PR #73, opened 2022-11-26
Improve environment variable handling

felixfontein:
Improve environment variable handling:
- Use the correct Sphinx markup for environment variables (`envvar` role).
- Create an environment variable index.

Examples:
- Index: https://ansible.fontein.de/collections/environment_variables.html
- Enviornment variable in a plugin: https://ansible.fontein.de/collections/ansible/builtin/sh_shell.html#ansible-collections-ansible-builtin-sh-shell (click it!)

### DIFF
--- a/changelogs/fragments/73-env-vars.yml
+++ b/changelogs/fragments/73-env-vars.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "Use correct markup (``envvar`` role) for environment variables. Compile an index of all environment variables used by plugins (https://github.com/ansible-community/antsibull-docs/pull/73)."
--- a/src/antsibull_docs/cli/doc_commands/stable.py
+++ b/src/antsibull_docs/cli/doc_commands/stable.py
@@ -37,6 +37,7 @@ from ...docs_parsing.routing import (
     load_all_collection_routing,
     remove_redirect_duplicates,
 )
+from ...env_variables import load_ansible_config, collect_referenced_environment_variables
 from ...schemas.docs import DOCS_SCHEMAS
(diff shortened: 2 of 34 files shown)


samccann (commented):
Tested locally and looks good except the new index of environement variables isn't part of any TOC that I can see? should be be in the list of 'index of xxx' like modules, vars plugins etc?

We should also add a link to this from here? https://docs.ansible.com/ansible/latest/reference_appendices/config.html#ansible-configuration-settings

felixfontein:
> Tested locally and looks good except the new index of environement variables isn't part of any TOC that I can see?

Yes, it would need to be added there somewhere.

> should be be in the list of 'index of xxx' like modules, vars plugins etc?

You mean in https://docs.ansible.com/ansible/devel/collections/all_plugins.html ? It would be a bit confusing if "Indexes of all modules and plugins" also contains an index of all environment variables. (On the other hand, that already contains an index of roles, but roles fit better in there than environment variables :) ) I'm not sure what the best place would be.

> We should also add a link to this from here? https://docs.ansible.com/ansible/latest/reference_appendices/config.html#ansible-configuration-settings

That requires a code change in ansible/ansible, but would be a good thing IMO.

samccann (approved):
docs and output LGTM

felixfontein:
Ok, plan is to move the 'env variable fallback documentation for modules' to a separate PR so the uncontroversial part can be merged first.

felixfontein:
I removed the module part (it is now a separate PR: #75).

felixfontein:
@samccann thanks for reviewing this!

PR #74, opened 2022-11-26
Automatically use short_description for seealso's description for modules and plugins when none is provided

### LINKED ISSUE #64: Display shortdescription for plugins listed in See Also section
Currenty, the See Also lists related plugins on a plugin docs page, but it doesn't display the short description for that plugin:

image

Would be great if we could replace that boilerplate text with the actual plugin short description.

felixfontein:
Fixes #64.

As an example, compare https://ansible.fontein.de/collections/ansible/builtin/add_host_module.html#see-also to https://docs.ansible.com/ansible/devel/collections/ansible/builtin/add_host_module.html#see-also.

### DIFF
--- a/src/antsibull_docs/augment_docs.py
+++ b/src/antsibull_docs/augment_docs.py
@@ -73,12 +73,13 @@ def _add_seealso(seealso: t.List[t.MutableMapping[str, t.Any]],
         elif entry.get('plugin') and entry.get('plugin_type'):
             plugin = entry['plugin']
             plugin_type = entry['plugin_type']
-        if (
-            plugin and plugin_type in plugin_info and plugin in plugin_info[plugin_type]
-            and isinstance(plugin_info[plugin_type][plugin].get('doc'), dict)
-            and plugin_info[plugin_type][plugin]['doc'].get('short_description')
-        ):
+        else:
+            continue
+        try:
             desc = plugin_info[plugin_type][plugin]['doc']['short_description']
+        except (KeyError, TypeError):
+            desc = None
+        if desc:
             if not desc.endswith(('.', '!', '?')):
                 desc += '.'
             entry['description'] = desc
--- a/src/antsibull_docs/augment_docs.py
+++ b/src/antsibull_docs/augment_docs.py
@@ -60,6 +60,30 @@ def add_full_key(options_data: t.Mapping[str, t.Any], suboption_entry: str,
                 _full_keys=full_keys_k)
 
 
+def _add_seealso(seealso: t.List[t.MutableMapping[str, t.Any]],
+                 plugin_info: t.Mapping[str, t.Mapping[str, t.Any]],
+                 ) -> None:
+    for entry in seealso:
+        if entry.get('description'):
+            continue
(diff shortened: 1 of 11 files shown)


src/antsibull_docs/augment_docs.py
```diff
@@ -60,6 +60,30 @@ def add_full_key(options_data: t.Mapping[str, t.Any], suboption_entry: str,
...
+            plugin = entry['module']
+        elif entry.get('plugin') and entry.get('plugin_type'):
+            plugin = entry['plugin']
+            plugin_type = entry['plugin_type']
+        if (
+            plugin and plugin_type in plugin_info and plugin in plugin_info[plugin_type]
+            and isinstance(plugin_info[plugin_type][plugin].get('doc'), dict)
+            and plugin_info[plugin_type][plugin]['doc'].get('short_description')
+        ):
+            desc = plugin_info[plugin_type][plugin]['doc']['short_description']
```
gotmax23, line 81: This feels a bit ugly to me. Would something like (

``` python
try:
    desc = plugin_info[plugin_type][plugin]['doc']['short_description']
except (KeyError, TypeError):
    continue
```

) be better?

src/antsibull_docs/augment_docs.py
```diff
@@ -60,6 +60,30 @@ def add_full_key(options_data: t.Mapping[str, t.Any], suboption_entry: str,
...
 
+def _add_seealso(seealso: t.List[t.MutableMapping[str, t.Any]],
+                 plugin_info: t.Mapping[str, t.Mapping[str, t.Any]],
+                 ) -> None:
+    for entry in seealso:
+        if entry.get('description'):
+            continue
+        plugin = ''
+        plugin_type = 'module'
+        if entry.get('module'):
```
gotmax23, line 71: Would `if ... in ...` work here and in the rest of this change? Or are you also trying to check if e.g. `entry['module']` is falsy when it's defined?

src/antsibull_docs/augment_docs.py
felixfontein: That would work, but then an additional `if desc:` would be needed to avoid `None` / `''` to be used.

src/antsibull_docs/augment_docs.py
felixfontein: I'm also checking for falsy when it's defined.

src/antsibull_docs/augment_docs.py
felixfontein: I've implemented that in af06c768fc88bb01d19b6192d544fa8cd19d6555.

gotmax23 (approved):
Thanks @felixfontein!

felixfontein:
@gotmax23 @samccann thanks for reviewing and testing!

src/antsibull_docs/augment_docs.py
felixfontein: This actually introduced a bug. The problem is that `plugin_info` can be a defaultdict, whence this construct modifies it when `plugin_type` is not present. #169 fixes this.

PR #75, opened 2022-12-01
[WIP] Allow module to improve environment variable handling

felixfontein:
Includes #73.

This is the controversal part of #73, which allows modules to declare environment variable fallbacks for options.

Related to https://github.com/ansible/ansible/pull/79471.


felixfontein:
Rebasing this exposed some problems with semantic markup and the environment variable overview page. I'll fix them next.

briantist:
The code so far looks good to me. I'm really looking forward to this!

felixfontein:
Core seems to really be opposed to this, so I'm not sure if this will ever be possible this way...

briantist:
> Core seems to really be opposed to this, so I'm not sure if this will ever be possible this way...

I don't see any real discussion in the core ticket :(
I don't really understand why they would be opposed. Modules do have env fallbacks whether they want them to or not, documenting them in a more visible way seems like a win.

PR #76, opened 2022-12-01
Implement Ansible version cut-off

felixfontein:
Stop mentioning when a feature was added in Ansible 2.6 or before.

For example, compare https://ansible.fontein.de/collections/community/sops/sops_encrypt_module.html#parameter-attributes with https://docs.ansible.com/ansible/devel/collections/community/sops/sops_encrypt_module.html#parameter-attributes, or https://ansible.fontein.de/collections/ansible/builtin/apt_module.html#parameter-only_upgrade with https://docs.ansible.com/ansible/devel/collections/ansible/builtin/apt_module.html#parameter-only_upgrade.

### DIFF
--- a/src/antsibull_docs/jinja2/tests.py
+++ b/src/antsibull_docs/jinja2/tests.py
@@ -24,13 +24,12 @@ TOO_OLD_TO_BE_NOTABLE = {
 test_list: t.Callable[[t.Any], bool] = partial(is_sequence, include_strings=False)
 
 
-def still_relevant(version, cutoff=None, collection=None):
+def still_relevant(version, collection=None):
     """
     Calculates whether the given version is older than a cutoff value
 
     :arg version: Version to check
-    :arg cutoff: Calculate whether `version` is older than this
-    :returns: True if the `version` is older than `cutoff` otherwise True.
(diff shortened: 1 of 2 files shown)


src/antsibull_docs/jinja2/tests.py
```diff
@@ -14,21 +14,22 @@
...
 
-def still_relevant(version, cutoff=TOO_OLD_TO_BE_NOTABLE, collection=None):
+def still_relevant(version, collection=None):
     """
     Calculates whether the given version is older than a cutoff value
 
     :arg version: Version to check
-    :arg cutoff: Calculate whether `version` is older than this
-    :returns: True if the `version` is older than `cutoff` otherwise True.
+    :returns: True if the `version` is older than the cutoff version, otherwise True.
```
samccann, line 32: is the end of this supposed to be 'otherwise False' ?

felixfontein: the `otherwise True` is correct, but the start should be `False`. Thanks for spotting this!

oraNod (approved):

felixfontein:
@samccann @oraNod thanks for reviewing this!

PR #80, opened 2022-12-06
Always write environment variable index; add it to index.rst generated by sphinx-init

felixfontein:
The updated index looks like this: https://ansible.fontein.de/ The new part is at the bottom.

This PR changes that the env var index is always written; in case it happens to be empty, `No environment variables have been defined.` will be shown.


samccann (approved):
Docs LGTM and thanks!

felixfontein:
@samccann thanks for reviewing this!

PR #83, opened 2022-12-15
Fix toctrees for collection's index page

felixfontein:
As reported by @evgeni on IRC.


evgeni:
This *looks* correct to me, but my Sphinx knowledge is a bit absent to do a full ack :)

felixfontein:
Restarting CI.

felixfontein:
I did test it with the foreman collection, and the result looked a lot better with this.

felixfontein:
@evgeni thanks for reporting and reviewing!

PR #92, opened 2023-01-02
sphinx-init: determine build script's path during runtime

### LINKED ISSUE #91: `sphinx-init` - paths in `build.sh` are not quoted, spaces lead to failures
When using `antsibull-docs sphinx-init` the generated `build.sh` is generated with a `cd` command into the full hardcoded directory name.

There are actually two issues with this I prefer to address:

1. If the path contains spaces, the script will fail because the paths are not quoted or escaped.
2. I would prefer the paths are not hardcoded, but relative.

For 1) we could quote or escape the path. Fixing 2) would fix 1) as well, but it can also be [deceptively complicated](https://stackoverflow.com/q/59895/3905079). 

One thing helping us with that maybe is the fact that the script is generated specifically to run in bash, so we don't really have to deal with all the edge cases for other shells.

FWIW, in my own uses, I replace that first line with this:
```
pushd "${BASH_SOURCE%/*}"
```
(and then I add `popd` to the end of the script)

We can use the above with `cd` as well, and optionally put a `cd -` at the end, though I don't think I've encountered a system where `pushd`/`popd` don't work, I think they are builtins.

---

At a minimum we should fix the quoting issue, not sure what the opinions are on the the relative path, but I like it for portability (I can move all the generated stuff around and have it still work).

felixfontein:
Fixes #91.


briantist (approved):
wow awesome! thanks @felixfontein !

felixfontein:

PR #94, opened 2023-01-11
Try to fix CI so we still have some tests running for Python 3.6

felixfontein:


felixfontein:
I'm looking forward when we can drop 3.6 support...

gotmax23:
I wouldn't mind dropping Poetry all together...

felixfontein:
I have to admit that I have start thinking that as well... I guess the main question is what to use instead...

PR #97, opened 2023-02-03
Fix pylint config

felixfontein:


felixfontein:
Oops, this PR contained more than it should have. Will revert....

PR #99, opened 2023-02-10
Explicitly declare and restrict sh dependency to < 2.0.0, and also declare other dependencies that is directly imported

felixfontein:


pyproject.toml
```diff
@@ -42,6 +42,12 @@ jinja2 = "*"
 packaging = "*"
 rstcheck = ">= 3.0.0, < 7.0.0"
 sphinx = "*"
+sh = ">= 1.0.0 < 2.0.0"
```
felixfontein, line 45:
```suggestion
# sh v2 has breaking changes.
# https://github.com/ansible-community/antsibull-core/issues/34
sh = ">= 1.0.0 < 2.0.0"
```

gotmax23: Thanks

pyproject.toml
```diff
@@ -42,6 +42,14 @@ jinja2 = "*"
 packaging = "*"
 rstcheck = ">= 3.0.0, < 7.0.0"
 sphinx = "*"
+# sh v2 has breaking changes.
+# https://github.com/ansible-community/antsibull-core/issues/34
+sh = ">= 1.0.0 < 2.0.0"
+pydantic = "*"
```
gotmax23, line 48: pydantic is the subject of a major rewrite. We might want to pin this to `< 2.0.0`.

gotmax23: In any case, this constraint should be the same between antsibull-docs and and antsibull-core.

gotmax23 (approved):

pyproject.toml
felixfontein: Sounds like a good idea. I'll create a follow-up for that (later today or tomorrow).

felixfontein:

pyproject.toml
felixfontein: One annoying thing is that pydantic does not adhere to semver, so `< 2.0.0` does not prevent breaking changes. But restricting this more is not a good idea either, IMO...

PR #112, opened 2023-03-26
Allow to validate Ansible markup

felixfontein:
Provides markup validation similar to https://github.com/ansible/ansible/pull/80243. Also allows to ensure that no semantic markup is used.


felixfontein:
I've tested this against all collections I have checked out locally (next to the ones I maintain: amazon.aws, ansible.netcommon, ansible.posix, ansible.util, community.aws, community.hashi_vault, community.mysql) and it worked fine. I'm going to merge this so I can continue with a harder problem: semantic markup for roles (which is unfortunately not part of the specs).

PR #113, opened 2023-03-26
Support semantic markup for roles

felixfontein:
Right now the semantic markup spec does not really support roles: these have entrypoints, which are not reflected in the `O()` and `RV()` syntaxes.

My proposal would be to use `:` for that as well:

- `O(foo)` inside a role's entrypoint refers to that entrypoint's `foo` otion;
- `O(main:foo)` inside a role's other entrypoint refers to the `foo` option of the `main` entrypoint;
- `O(main:foo)` outside a role's entrypoint, but inside the role (does not really exist yet) refers to the `main` entrypoint's `foo` option;
- `O(foo.bar.baz#role:main:foo)` refers to the `foo` option of the `main` entrypoint of the `foo.bar.baz` role.

(Roles do not have return value documentation yet, but I've already done the same changes there in case they are ever added.)

An example PR of how this looks for roles in a collection: https://github.com/felixfontein/ansible-acme/pull/50
The results can be seen here:
- https://ansible.fontein.de/_felixfontein.acme/revoke_old_certificates_role.html
- https://ansible.fontein.de/_felixfontein.acme/acme_certificate_role.html

Compare them to the same docs without semantic markup:
- https://ansible.fontein.de/collections/felixfontein/acme/revoke_old_certificates_role.html
- https://ansible.fontein.de/collections/felixfontein/acme/acme_certificate_role.html


briantist:
I like this! 

Minor point: I would also suggest to allow the partially- or fully-qualified name format even inside an entrypoint. Maybe it already works that way, but from the description it's not certain:

> O(main:foo) inside a role's other entrypoint

For example, this says inside a role's _other_ entrypoint, but imo it should also work inside the `main` entrypoint if used that way.

> O(foo.bar.baz#role:main:foo) refers to the foo option of the main entrypoint of the foo.bar.baz role.

Similarly, I hope this would also work inside the `foo.bar.baz` role's `main` entrypoint as well.

Hope that makes sense!

felixfontein:
Yes, the longer forms already work.

felixfontein:
The box around the box happens if the link is broken, which it turned out to be in this case. I found a bug in the markup validator and several bugs in https://github.com/felixfontein/ansible-acme/pull/50 :)

felixfontein:
@briantist @samccann thanks for your reviews and comments!

PR #115, opened 2023-04-05
Switch to hatchling, and requiring Python 3.9+

felixfontein:


gotmax23 (commented):
Here are some initial comments

.github/workflows/antsibull-docs.yml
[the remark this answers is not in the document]
felixfontein: I first wanted to do that, but than decided to not do it in this PR since it is a quite extensive job, and also there would be five different targets and not just one (you really want to run them in parallel since they take a long time).

pyproject.toml
```diff
@@ -3,78 +3,129 @@
...
+packages = ["src/antsibull_docs", "src/sphinx_antsibull_ext"]
 
 [tool.isort]
-line_length = 100
+# Match black
+profile = "black"
+line_length = 89
+#
 multi_line_output = 3
 include_trailing_comma = true
```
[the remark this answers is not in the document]
felixfontein: I prefer the "black" import style over the default isort one, which IMO is ugly.

noxfile.py
```diff
@@ -0,0 +1,314 @@
...
+        "--cov-report",
+        "term-missing",
+        *more_args,
+        *session.posargs,
+        env={"COVERAGE_FILE": f"{covfile}", **os.environ},
+    )
+
+
+@nox.session
+def coverage_release(session: nox.Session):
```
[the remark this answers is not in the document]
felixfontein: Yes. I first kept it when copying the noxfile over from antsibull as I wanted to replace it with the integration test part from .github/workflows/antsibull-docs.yml, but then after deciding to not do that in this PR I forgot about this part (and the corresponding part in the README)...

.github/workflows/antsibull-docs.yml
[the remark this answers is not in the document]
gotmax23: nox supports [parametrizing sessions](https://nox.thea.codes/en/stable/config.html#parametrizing-sessions) which would make this easier, but yeah, I think it's reasonable to do that separately.

noxfile.py
```diff
@@ -0,0 +1,267 @@
...
+        "git",
+        "tag",
+        "-a",
+        "-m",
+        f"antsibull-docs {version}",
+        "--edit",
+        version,
+        external=True,
+    )
+    session.run("hatch", "build")
```
gotmax23, line 238: This will clean existing artifacts before building. I think we should start using this in the other projects, also. If there's other artifacts in the `dist` directory (e.g. a test build for a `.post0` version that shouldn't be released), `hatch publish` will publish it.

```suggestion
    session.run("hatch", "build", "--clean")
```

.github/workflows/antsibull-docs.yml
```diff
@@ -59,28 +59,25 @@ jobs:
...
+          sed -i build.sh -e 's!sphinx-build !coverage run -p --source antsibull_docs --source sphinx_antsibull_ext -m sphinx.cmd.build !g'
           cat build.sh
         working-directory: antsibull-docs
 
       - name: Install dependencies
         run: |
-          poetry run pip install ansible-core
-          poetry run pip install -r requirements.txt
+          pip install ansible-core
+          pip install -r requirements.txt
```
gotmax23, line 80:
```suggestion
          pip install ansible-core -r requirements.txt
```

.github/workflows/antsibull-docs.yml
```diff
@@ -59,28 +59,25 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install poetry
-          sed -i -e 's/^python = .*/python = "^${{ matrix.python }}"/' pyproject.toml
-          poetry install
-          poetry update
+          pip install -e .[coverage] ../antsibull-core
```
gotmax23, line 62:
```suggestion
          pip install .[coverage] ../antsibull-core
```

.github/workflows/nox.yml
```diff
@@ -0,0 +1,72 @@
...
+    steps:
+      - name: Check out antsibull-docs
+        uses: actions/checkout@v3
+        with:
+          path: antsibull-docs
+      - name: Check out dependent project antsibull-core
+        uses: actions/checkout@v3
+        with:
+          repository: ansible-community/antsibull-core
+          path: antsibull-core
```
gotmax23, line 37: You can remove this if you'd like, as nox already handles installing antsibull-core from `main`.

gotmax23 (commented):
I have a couple last comments. Other than that, I think we're good to go :smile:.

.github/workflows/antsibull-docs.yml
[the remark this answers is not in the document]
felixfontein: IMO they make the workflow more complicated, and if the plan is to get rid of that workflow anyway and replace it with nox, then I don't think it's worth investing additional work here.

noxfile.py
felixfontein: Yeah, I noticed that it tries to upload everything it finds in `dist/`. Cleaning up is thus :+1:.

.github/workflows/antsibull-docs.yml
felixfontein: I prefer editable installs when doing code coverage. Otherwise coverage reports could be a mess since random paths are in there.

.github/workflows/nox.yml
felixfontein: I think it's easier to leave it here so if we ever need to check out a specific branch (instead of `main`), it's a one-line change here, instead of adding this whole block first.

pyproject.toml
```diff
@@ -3,78 +3,129 @@
...
-# pydantic v2 is a major rewrite
-pydantic = ">= 1.0.0 < 2.0.0"
-semantic_version = "*"
-aiohttp = ">= 3.0.0"
-twiggy = "*"
-PyYAML = "*"
+[project.optional-dependencies]
+codeqa = [
+    "flake8 >= 3.8.0",
+    "pylint ~= 2.17.2",
```
felixfontein, line 73:
```suggestion
    "pylint >= 2.17.2",
```

pyproject.toml
```diff
@@ -3,78 +3,129 @@
...
+    "cryptography",
+    "pytest",
+    "pytest-asyncio >= 0.12",
+    "pytest-cov",
+    "pytest-error-for-skips",
+]
+typing = [
+    "mypy",
+    # https://github.com/facebook/pyre-check/issues/398
+    "pyre-check ~= 0.9.17",
```
felixfontein, line 91:
```suggestion
    "pyre-check >= 0.9.17",
```

.github/workflows/nox.yml
gotmax23: I was also thinking that.

/resolve

gotmax23 (approved):
Ack. Thanks for working on this!

felixfontein:

PR #116, opened 2023-04-06
Speed up functional tests

felixfontein:


felixfontein:
The build and lint tests each call `ansible --version` and `ansible-galaxy collection list` as well. This is something we can also cache to speed this up even further.

PR #118, opened 2023-04-08
Clean up code by moving functions to better places

felixfontein:


src/antsibull_docs/cli/doc_commands/_build.py
```diff
@@ -0,0 +1,217 @@
...
+"""Utilities for various docs build subcommands."""
+
+import textwrap
+import typing as t
+
+from antsibull_core.compat import asyncio_run
+from antsibull_core.logging import log
+from antsibull_core.venv import FakeVenvRunner, VenvRunner
+
+from ... import app_context
```
gotmax23, line 15: I find these deep relative imports very confusing. IMO, using the absolute name would be better.

felixfontein: Hmm, I personally like that we always use relative imports for things in the same antsibull* package.

PR #120, opened 2023-04-09
Drop support for ansible-core 2.12 and earlier

felixfontein:
Allows to remove quite some code for docs parsing, and allows to use `ansible-galaxy collection list --format json` instead of parsing its text output.


gotmax23 (approved):
LGTM besides that one stray comment.

felixfontein:
Note to self: antsibull-core 2.0.0 can likely remove docs-specific options from the default context.

felixfontein:

PR #121, opened 2023-04-09
Allow to use currently installed ansible-core for devel and stable subcommands

felixfontein:
CC @mattclay

### DIFF
--- a/src/antsibull_docs/cli/antsibull_docs.py
+++ b/src/antsibull_docs/cli/antsibull_docs.py
@@ -280,7 +280,8 @@ def parse_args(program_name: str, args: List[str]) -> argparse.Namespace:
     devel_parser.add_argument('--use-installed-ansible-core', action='store_true',
                               help='Assumes that ansible-core is already installed and can be'
                               ' used by calling `ansible`, `ansible-doc`, and `ansible-galaxy`'
-                              ' from $PATH.')
+                              ' from $PATH.'
+                              ' By default, antsibull-docs installs ansible-core into a temporary venv.')
 
     #
     # Document a released version of ansible
--- a/src/antsibull_docs/cli/antsibull_docs.py
+++ b/src/antsibull_docs/cli/antsibull_docs.py
@@ -280,8 +280,8 @@ def parse_args(program_name: str, args: List[str]) -> argparse.Namespace:
     devel_parser.add_argument('--use-installed-ansible-core', action='store_true',
                               help='Assumes that ansible-core is already installed and can be'
                               ' used by calling `ansible`, `ansible-doc`, and `ansible-galaxy`'
-                              ' from $PATH.'
-                              ' By default, antsibull-docs installs ansible-core into a temporary venv.')
+                              ' from $PATH. By default, antsibull-docs installs ansible-core'
+                              ' into a temporary venv.')
 
     #
     # Document a released version of ansible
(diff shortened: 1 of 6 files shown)


src/antsibull_docs/cli/antsibull_docs.py
```diff
@@ -277,6 +277,10 @@ def parse_args(program_name: str, args: List[str]) -> argparse.Namespace:
                                          ' release of Ansible')
     devel_parser.add_argument('--pieces-file', default=DEFAULT_PIECES_FILE,
                               help='File containing a list of collections to include')
+    devel_parser.add_argument('--use-current-ansible-core', action='store_true',
```
gotmax23, line 280: IMO, something like `--use-installed-ansible-core` would be better. `current` seems confusing here.

felixfontein: Renamed in ab51201a32d2cfa3bd911afeab8ee7a9b82aa26b.

src/antsibull_docs/cli/antsibull_docs.py
```diff
@@ -277,6 +277,10 @@ def parse_args(program_name: str, args: List[str]) -> argparse.Namespace:
                                          ' release of Ansible')
     devel_parser.add_argument('--pieces-file', default=DEFAULT_PIECES_FILE,
                               help='File containing a list of collections to include')
+    devel_parser.add_argument('--use-installed-ansible-core', action='store_true',
+                              help='Assumes that ansible-core is already installed and can be'
+                              ' used by calling `ansible`, `ansible-doc`, and `ansible-galaxy`'
+                              ' from $PATH.')
```
gotmax23, line 283:
```suggestion
                              help='Assumes that ansible-core is already installed and can be'
                              ' used by calling `ansible`, `ansible-doc`, and `ansible-galaxy`'
                              ' from $PATH.'
                              ' By default, antsibull-docs installs ansible-core into a temporary venv.')
```

src/antsibull_docs/cli/antsibull_docs.py
```diff
@@ -289,6 +293,10 @@ def parse_args(program_name: str, args: List[str]) -> argparse.Namespace:
     stable_parser.add_argument('--deps-file', required=True,
                                help='File which contains the list of collections and'
                                ' versions which were included in this version of Ansible')
+    stable_parser.add_argument('--use-installed-ansible-core', action='store_true',
+                               help='Assumes that ansible-core is already installed and can be'
+                               ' used by calling `ansible`, `ansible-doc`, and `ansible-galaxy`'
+                               ' from $PATH.')
```
gotmax23, line 299:
```suggestion
                              help='Assumes that ansible-core is already installed and can be'
                              ' used by calling `ansible`, `ansible-doc`, and `ansible-galaxy`'
                              ' from $PATH.'
                              ' By default, antsibull-docs installs ansible-core into a temporary venv.')
```

gotmax23 (approved):
One other small docs suggestion. Otherwise, this looks good to me. Thanks!

mattclay (changes requested):
It seems this PR is still missing one necessary change. It doesn't look like it provides a way to preserve `PYTHONPATH`, which is being removed here:

https://github.com/ansible-community/antsibull-docs/blob/489da646f9784796a75e06426831e5c8f3f4b412/src/antsibull_docs/docs_parsing/__init__.py#L34

This prevents antsibull-docs from recognizing the ansible-core that is injected into the virtual environment.

mattclay (approved):

felixfontein:
@gotmax23 @mattclay thanks for reviewing this!

mattclay:
@felixfontein Do you have an idea of when you'll be releasing a version with this change?

felixfontein:
@mattclay first antsibull-core 2.0.0 needs to be released before antsibull-docs 2.0.0 can be released (which will contain this change). We're in the middle of preparing both; I'm right now not sure how long this will take. I hope/think/assume both will be done in a couple of weeks, but it depends a lot on when/how much @gotmax23 and me have time to work on it...

gotmax23:
Is this worth backporting?

felixfontein:
It's a feature, so no backport. `stable-1` is reserved for bugfixes (so we can update antsibull-docs in ansible-core 2.15).

PR #122, opened 2023-04-10
Remove sh usages

felixfontein:
Needs to be rebased/updated once #120 has been merged.

### DIFF
--- a/changelogs/fragments/122-antsibull-core-sh.yml
+++ b/changelogs/fragments/122-antsibull-core-sh.yml
@@ -1,2 +1,3 @@
 minor_changes:
   - "Now depends antsibull-core 2.0.0 or newer; antsibull-core 1.x.y is no longer supported (https://github.com/ansible-community/antsibull-docs/pull/122)."
+  - "Ansibull-docs now no longer depends directly on ``sh`` (https://github.com/ansible-community/antsibull-docs/pull/122)."
--- a/changelogs/fragments/122-antsibull-core-sh.yml
+++ b/changelogs/fragments/122-antsibull-core-sh.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "Now depends antsibull-core 2.0.0 or newer; antsibull-core 1.x.y is no longer supported (https://github.com/ansible-community/antsibull-docs/pull/122)."
--- a/src/antsibull_docs/docs_parsing/ansible_doc.py
+++ b/src/antsibull_docs/docs_parsing/ansible_doc.py
@@ -8,10 +8,10 @@
 import json
 import os
 import re
-import subprocess
 import typing as t
 
 from antsibull_core.logging import log
+from antsibull_core.subprocess_util import CalledProcessError
 from antsibull_core.vendored.json_utils import _filter_non_json_lines
 from packaging.version import Version as PypiVer
 
@@ -123,7 +123,7 @@ async def get_ansible_core_version(venv: t.Union['VenvRunner', 'FakeVenvRunner']
         if metadata.version is None:
             raise ValueError('Cannot retrieve ansible-core version from `ansible --version`')
         return PypiVer(metadata.version)
-    except subprocess.CalledProcessError as exc:
+    except CalledProcessError as exc:
         raise ValueError(
             f'Cannot retrieve ansible-core version from `ansible --version`: {exc}'
         ) from exc
--- a/src/antsibull_docs/docs_parsing/ansible_doc.py
+++ b/src/antsibull_docs/docs_parsing/ansible_doc.py
@@ -9,11 +9,8 @@ import json
 import os
 import re
 import subprocess
-import sys
-import traceback
 import typing as t
 
-import sh
 from antsibull_core.logging import log
 from antsibull_core.vendored.json_utils import _filter_non_json_lines
 from packaging.version import Version as PypiVer
@@ -78,7 +75,7 @@ async def _call_ansible_version(
 async def _call_ansible_galaxy_collection_list(
     venv: t.Union['VenvRunner', 'FakeVenvRunner'],
     env: t.Dict[str, str],
-) -> str:
+) -> t.Mapping[str, t.Any]:
     p = await venv.async_log_run(
         ['ansible-galaxy', 'collection', 'list', '--format', 'json'],
         env=env,
--- a/src/antsibull_docs/docs_parsing/ansible_doc.py
+++ b/src/antsibull_docs/docs_parsing/ansible_doc.py
@@ -8,6 +8,9 @@
 import json
 import os
 import re
+import subprocess
+import sys
+import traceback
 import typing as t
 
 import sh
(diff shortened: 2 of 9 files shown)


felixfontein:
@gotmax23 the failing tests show a problem with `ansible_core.subprocess_util.async_log_run`: if a single line is too large of the output (in this case stdout), asyncio.streams dies:
```
[...]
  File "/opt/hostedtoolcache/Python/3.11.2/x64/lib/python3.11/site-packages/antsibull_core/subprocess_util.py", line 40, in _stream_log
    line = await stream.readline()
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.11.2/x64/lib/python3.11/asyncio/streams.py", line 554, in readline
    raise ValueError(e.args[0])
ValueError: Separator is found, but chunk is longer than limit
```
This requires a very large output, which apparently happens for `ansible-doc --metadata-dump --no-fail-on-errors` when the `ansible` package is installed. For the 8.0.0 build I did recently, the maximum line length I found was 68115.

According to the asyncio/streams.py in Python 3.10, the default limit is 65536 bytes for a line. This can be adjusted by passing `limit=HIGHER_LIMIT` to `asyncio.create_subprocess_exec()`, but IMO this is only a band-aid, since there is no value for "unlimited". We might have to implement line parsing ourselves?

felixfontein:
I've tried it out in https://github.com/ansible-community/antsibull-core/pull/53, it seems to work well.

felixfontein:
Restarting tests now that https://github.com/ansible-community/antsibull-core/pull/53 has been merged.

felixfontein:
(Rebased to remove conflict.)

gotmax23:
Thanks for working on this! I'll take a closer look later today or tomorrow.

src/antsibull_docs/lint_plugin_docs.py
```diff
@@ -74,8 +74,8 @@ def __exit__(self, type_, value, traceback_):
 class CollectionFinder:
     def __init__(self):
         self.collections = {}
-        stdout = sh.Command('ansible-galaxy')('collection', 'list', '--format', 'json').stdout
-        raw_output = stdout.decode('utf-8', errors='surrogateescape')
+        p = log_run(['ansible-galaxy', 'collection', 'list', '--format', 'json'])
+        raw_output = p.stdout.decode('utf-8', errors='surrogateescape')
         data = json.loads(_filter_non_json_lines(raw_output)[0])
```
gotmax23, line 79: `p.stdout` is already a string.

```suggestion
        p = log_run(['ansible-galaxy', 'collection', 'list', '--format', 'json'])
        data = json.loads(_filter_non_json_lines(p.stdout)[0])
```

changelogs/fragments/122-antsibull-core-sh.yml
gotmax23, line 1: I'd also mention that we no longer (directly) depend on `sh`.

gotmax23 (commented):
This PR is pretty straightfoward; I just have a couple comments :smile:.

src/antsibull_docs/docs_parsing/ansible_doc.py
```diff
@@ -105,28 +104,26 @@ def get_collection_metadata(venv: t.Union['VenvRunner', 'FakeVenvRunner'],
...
-    try:
-        venv_python = venv.get_command('python')
-        ansible_version_cmd = venv_python(
-            '-c', 'import ansible.release; print(ansible.release.__version__)', _env=env)
-        output = ansible_version_cmd.stdout.decode('utf-8', errors='surrogateescape').strip()
+async def get_ansible_core_version(venv: t.Union['VenvRunner', 'FakeVenvRunner'],
+                                   env: t.Optional[t.Dict[str, str]] = None,
+                                   ) -> PypiVer:
+    p = await venv.async_log_run(
+        ['python', '-c', 'import ansible.release; print(ansible.release.__version__)'],
```
gotmax23, line 111: I opened https://github.com/ansible-community/antsibull-core/pull/65.

src/antsibull_docs/docs_parsing/ansible_doc.py
```diff
@@ -64,38 +64,37 @@ def parse_ansible_galaxy_collection_list(json_output: t.Mapping[str, t.Any],
...
-    ansible_version_cmd = venv_ansible('--version', _env=env)
-    return ansible_version_cmd.stdout.decode('utf-8', errors='surrogateescape')
+    p = await venv.async_log_run(['ansible', '--version'], env=env)
+    return p.stdout
 
 
-def _call_ansible_galaxy_collection_list(
+async def _call_ansible_galaxy_collection_list(
     venv: t.Union['VenvRunner', 'FakeVenvRunner'],
     env: t.Dict[str, str],
```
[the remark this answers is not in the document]
felixfontein: The only one that accepts `None` is `_call_ansible_version()`. Changing all to `t.Optional[]` would simply make the tests code more complicated (since then all three mock functions must handle `None`, and not just one of them), but have no other advantage from my POV.

src/antsibull_docs/lint_plugin_docs.py
felixfontein: Right...

felixfontein: 8b35cf4c6d40caec7c62806369cb6a65407538ba

changelogs/fragments/122-antsibull-core-sh.yml
felixfontein: ff4426c086617272c9c1af2f1e501879f2094f83

src/antsibull_docs/docs_parsing/ansible_doc.py
felixfontein: That would be a breaking change for this code, this one is actively running the first `python` in path instead of using `sys.executable` for the fake venv.

felixfontein: I thought a bit more about this (and looked again at the code), and I think this change would be fine. Calling `python -c '...'` is a short-cut to avoid the slow start-up time of `ansible --version`, and if it fails (because the wrong Python is used) it will fall back to `ansible --version` anyway.

gotmax23: Right. Although, now that I'm also thinking... is there a reason we don't just run `ansible.release; print(ansible.release.__version__)` directly? It seems strange to me to shell out to `python -c` from python, but maybe there's some historical reason to this.

felixfontein: If the venv is not a fake venv, but a real one, you cannot simply directly import Python code from it.

gotmax23: :facepalm: right....

gotmax23 (approved):
Ack. Let's do this :tada:!

felixfontein:
@gotmax23 thanks for preparing this and reviewing!

PR #130, opened 2023-04-19
Use modern typing syntax and run isort

gotmax23:
Update to the modern type annotation syntax instead of using the generics in the `typing` module. All of the files except those that import pydantic use `from __future__ import annotations`.

As part of this change, I ran isort over the whole codebase.

### DIFF
--- a/src/antsibull_docs/markup/_counter.py
+++ b/src/antsibull_docs/markup/_counter.py
@@ -7,13 +7,15 @@
 Count markup instructions in parsed markup.
 """
 
-import typing as t
+from __future__ import annotations
+
+from collections.abc import Sequence
 
 from antsibull_docs_parser import dom
 
 
 class _Counter(dom.Walker):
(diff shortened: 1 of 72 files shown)


src/antsibull_docs/markup/_counter.py
```diff
@@ -78,7 +80,7 @@ def process_return_value(self, part: dom.ReturnValuePart) -> None:
         self.counts['return-value'] += 1
 
 
-def count(paragraphs: t.Sequence[dom.Paragraph]) -> t.Dict[str, int]:
+def count(paragraphs: t.Sequence[dom.Paragraph]) -> dict[str, int]:
```
gotmax23, line 83: TODO: Replace this with `collections.abc.Sequence`

felixfontein:
This now has conflicts.

felixfontein (approved):

felixfontein:
Thanks a lot!

Issue #31, opened 2022-08-31
Render changelogs in docsite build


felixfontein:
Inspired by https://github.com/ansible-collections/community.hashi_vault/pull/299

There are a couple of things to discuss/decide:

1. Should this be done for every collection that has a changelog in the correct format, or only for collections who opt-in?

2. Should it use changelogs/changelog.yaml as the source of truth, or some rendered CHANGELOG.rst file? The trouble with the former (which I would prefer) is that antsibull-docs then needs to depend on antsibull-changelog in some form, which potentially leads back to the trouble of "updating antsibull-docs forces an update of antsibull-changelog" in ansible-core's test containers (antsibull-docs is only needed in the ansible-core default test container, while antsibull-changelog is included in all default test containers and thus updates to it are potentially user noticeable).

mariolenz:
I think a changelog is an important part of the documentation, so I tend to have have it in docsite for all collections that have one in the correct format. This way, all users would benefit from having the changelog alongside the module documentation without the collection maintainers having to do something.

But I could also live with opt-in.

briantist:
1. I would prefer that all collections which have one in the proper format have it in the docs, unless they opt-out of it.

I had used the "opt-in" language before, but for me this was more about letting collections use a different method publishing a changelog if they wish. But the "if they have [one] in the proper format" covers that well enough. I suppose we could allow an opt-out just in case they have one in the right format, but for some reason don't want it?

I especially like that having it "default on" (opt-out) means that most collections in the ansible package, if not all, and most collections in general that use the antsibull tools, will have the changelog viewable in docs, with no additional action needed.

I could also live with opt-in, but it can be hard to bring 100+ collections into changes like that, so adoption might be less.

---

2. I agree that using the YAML version as the source of truth sounds better. But the dependency thing sounds tricky. 

What I might suggest then is starting out by using a rendered changelog for this feature. I also like the idea of being able to configure the path to the changelog, while it can default to `<root>/CHANGELOG.rst`. That would actually add a little flexibility for anyone who either renders a changelog with a different process (as long as it's RST in the end), no matter the name or location.

Using rendered also allows for a possibility of the rendered file being edited though I think that will be rare in the antsibull workflow since changes would be overwritten.

As a _optional_ follow-up feature, if the dependency issues could be worked out, a YAML-based form could be introduced. If so, this could become the default mode (when nothing is configured), though for some of the reasons above, I think having an option to point to a rendered file is still good.

[...]

PR #134, opened 2023-04-23
Disable flatmapping except for community.general < 6.0.0 and community.network < 5.0.0

### LINKED ISSUE #132: Bad name for module in custom collection
With my team, we have a custom collection. In this collection we have custom modules. 

In particular, we have these directory layout : 

image

These modules must be called in our playbook with the entire directory names. For example, module **plugins/modules/cde/job.py** is called with **namespace.collection_name.<i>cde</i>.job**.

We're also building our documentation through **antsibull-docs**. 

Doing so we have a website with all our module documentation. However the name if the module name on the docs is not correctly set : in particular, we miss the intermediate directory name. For example, module **plugins/modules/cde/job.py** is referenced under **namespace.collection_name.job** (so without the ***cde*** directory). 

image

In conclusion, I think that there is an issue when the docs is created : it looks like it removes all intermediate directories in the module names.

felixfontein:
Fixes #132.

I'm not sure whether it is worth adding an option to allow to enable it for specific collections. So for now, I've disabled it for all (except for c.g < 6.0.0 and c.n < 5.0.0).


gotmax23:
IMO, we should either deprecate this functionality ("deprecation/feature removal: support for flatmapping has been removed. For now, it is kept as a special case for community.general < 6.0.0 and community.network <5.0.0") or make this configurable ("breaking change: this option is now disabled by default. As a special case, it is always enabled for community.general < 6.0.0 and community.network < 5.0.0")

felixfontein:
I would like to make it configurable, but the **how** is the largest problem. This needs to be a per-collection configuration, but right now there is no way for a collection to say "please flatmap me".

felixfontein:
Regarding configuring:
1. We cannot use MANIFEST.json / galaxy.yml.
2. Simply using `changelogs/config.yaml`'s value `flatmap` is kind of strange, since `changelogs/config.yaml` isn't really a config file for antsibull-docs.
3. Adding a new option to `docs/docsite/extra-docs.yml` or `docs/docsite/links.yml` seems even worse.
4. We could add a new file `docs/docsite/config.yml`. That could also be used for #31.
5. Waiting for something like https://github.com/ansible-community/community-topics/issues/154#issuecomment-1464590465 will probably take too long.

I'm currently tending to 4. WDYT?

felixfontein:
Ok, I've now implemented 4.

gotmax23 (approved):
Ack

felixfontein:
@gotmax23 thanks a lot for reviewing this!

PR #135, opened 2023-04-23
Fix wrong typings

felixfontein:
For some reason the `retrieve()` functions claimed to return versions and not paths. And for some other (unknown) reasons, the type checkers didn't complain that a list of versions was passed when a list of strings was expected.

I suddenly started to get error messages locally that I didn't see in CI:
```
src/antsibull_docs/cli/doc_commands/collection.py:141:37 Incompatible parameter type [6]: In call `install_together`, for 1st positional argument, expected `List[str]` but got `List[Version]`.
src/antsibull_docs/cli/doc_commands/devel.py:144:37 Incompatible parameter type [6]: In call `install_together`, for 1st positional argument, expected `List[str]` but got `List[Version]`.
src/antsibull_docs/cli/doc_commands/devel.py:153:33 Incompatible parameter type [6]: In call `VenvRunner.install_package`, for 1st positional argument, expected `str` but got `Version`.
src/antsibull_docs/cli/doc_commands/stable.py:146:37 Incompatible parameter type [6]: In call `install_together`, for 1st positional argument, expected `List[str]` but got `List[Version]`.
src/antsibull_docs/cli/doc_commands/stable.py:155:33 Incompatible parameter type [6]: In call `VenvRunner.install_package`, for 1st positional argument, expected `str` but got `Version`.
```

### DIFF
--- a/src/antsibull_docs/cli/doc_commands/collection.py
+++ b/src/antsibull_docs/cli/doc_commands/collection.py
@@ -12,7 +12,6 @@ import asyncio
 import os
 import os.path
 import tempfile
-import typing as t
 
 import aiohttp
 import asyncio_pool  # type: ignore[import]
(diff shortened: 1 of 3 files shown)


gotmax23 (approved):
Ack.

PR #137, opened 2023-04-24
nox typing: always install self in editable mode

gotmax23:
pyre does not work when we don't install ourself in editable mode.


gotmax23:
I'm triggering CI. Now that I've merged https://github.com/ansible-community/antsibull-docs/pull/135, this should pass.

felixfontein:
Thanks for figuring this out!

PR #145, opened 2023-05-06
Reformat with black

felixfontein:
Must not be merged, but pushed directly to `main`.


pyproject.toml
```diff
@@ -108,10 +109,15 @@ dev = [
...
+[tool.black]
+exclude = '''
+tests/functional/baseline-sphinx-init-.*
+'''
+
 [tool.isort]
 # Match black
 profile = "black"
-line_length = 89
+line_length = 88
```
felixfontein, line 120: Without this, isort reformats a line that black instantly changes back.

gotmax23: Hmm, I should have realized that before I copied it to all the antsibull projects... Thanks for catching it :).

Looking at https://pycqa.github.io/isort/docs/configuration/profiles.html, I think setting `profile = "black"` and removing the other options is a better choice. The black profile includes everything else that is set here.

felixfontein: Ah, I didn't knew that 88 was the 'right' value. I simply decreased 89 by one and then the modify-and-back cycle didn't happen anymore...

.flake8
```diff
@@ -8,5 +8,5 @@ extend-ignore = E203
 count = true
 max-complexity = 10
 # black's max-line-length is 89, but it doesn't touch long string literals.
-max-line-length = 100
+max-line-length = 120
```
gotmax23, line 11: Wouldn't it make more sense to shorten the one overly long line in `src/antsibull_docs/write_docs/plugins.py:171` instead of raising max-line-length to 120?

felixfontein: It would make more sense if black would stop artifically making this line longer than necessary by changing
```
        edit_on_github_url = (
            f"https://github.com/{eog.repository}/edit/{eog.branch}/{eog.path_prefix}{gh_path}"
        )
```
(less than 100 chars) back to 
```
        edit_on_github_url = f"https://github.com/{eog.repository}/edit/{eog.branch}/{eog.path_prefix}{gh_path}"
```
I've now split it up into even more lines so black won't screw it up.

gotmax23 (approved):
Ack. Thanks for working on this!

felixfontein:

PR #146, opened 2023-05-07
Add example config file

felixfontein:
I noticed I've never checked it in...

### DIFF
--- a/antsibull-docs.cfg
+++ b/antsibull-docs.cfg
@@ -1,0 +1,35 @@
+# Copyright (c) Ansible Project
+# GNU General Public License v3.0+ (see LICENSES/GPL-3.0-or-later.txt or https://www.gnu.org/licenses/gpl-3.0.txt)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+# Determines whether to add breadcrumbs to plugin docs via hidden sphinx toctrees. This can
+# take up a significant amount of memory if there are a large number of plugins so you can
+# disable this if necessary.
+breadcrumbs = true
+
+# Which docs parsing backend to use. Currently there is only 'ansible-core-2.13', and 'auto'
+# always chooses that one.
+doc_parsing_backend = "auto"
+
+# Determines whether to create the collection index and plugin indexes. They may not be
+# needed if you have a different structure for your website.
+indexes = true
+
+# Determines whether to use HTML blobs for option and return value tables. Using HTML blobs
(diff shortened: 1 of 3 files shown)


antsibull-docs.cfg
```diff
@@ -0,0 +1,35 @@
...
+# Which docs parsing backend to use. Currently there is only 'ansible-core-2.13', and 'auto'
+# always chooses that one.
+doc_parsing_backend = "auto"
+
+# Determines whether to create the collection index and plugin indexes. They may not be
+# needed if you have a different structure for your website.
+indexes = true
+
+# Determines whether to use HTML blobs for option and return value tables. Using HTML blobs
+# reduces memory and CPU time usage significantly so you can disable this if necessary.
```
gotmax23, line 19: Should this be

```suggestion
# reduces memory and CPU time usage significantly so you can enable this if necessary.
```

?

antsibull-docs.cfg
```diff
@@ -0,0 +1,35 @@
...
+use_html_blobs = false
+
+# You can specify ways to convert a collection name (<namespace>.<name>) to an URL here.
+# You can replace either of <namespace> or <name> by "*" to match all values in that place,
+# or use "*" for the collection name to match all collections. In the URL, you can use
+# {namespace} and {name} for the two components of the collection name. If you want to use
+# "{" or "}" in the URL, write "{{" or "}}" instead. Basically these are Python format
+# strings (https://docs.python.org/3.8/library/string.html#formatstrings).
+collection_url = {
+  * = "https://galaxy.ansible.com/{namespace}/{name}"
```
gotmax23, line 29: Note to self: I think we'll need to change this when Galaxy v3 is deployed.

gotmax23 (approved):

antsibull-docs.cfg
felixfontein: Indeed! There should be compatibility redirects for existing collections, so it won't break the docsite for older stable branches, but once new collections are added this can become problematic.

antsibull-docs.cfg
felixfontein: Good spot. I copied this from the CLI argument help, so I also fixed that one in 8f6f4038f460dbcdb14168e3f018f8003a167d9c.

antsibull-docs.cfg
felixfontein: I created an issue for this: #147

felixfontein:

gotmax23:
Thank you!

PR #154, opened 2023-06-01
Setting default contents of the description key of the role endpoint schema

jpodivin:
Existing ansible role documentation doesn't imply that the key is required. In fact the examples seem to indicate the opposite [0].

This is potentially confusing to developers working on documenting their ansible collections.

The proposed change should simplify implementing new argument specs. 

[0] https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html#sample-specification

### DIFF
--- a/src/antsibull_docs/schemas/docs/role.py
+++ b/src/antsibull_docs/schemas/docs/role.py
@@ -50,7 +50,7 @@ class RoleOptionsSchema(OptionsSchema):
 class RoleEntrypointSchema(BaseModel):
     """Documentation for role entrypoints."""
 
-    description: list[str]
+    description: list[str] = []
     short_description: str
     author: list[str] = []
     deprecated: DeprecationSchema = p.Field({})


felixfontein:
I don't think it is a good idea to allow empty descriptions.

felixfontein:
Ref: https://github.com/ansible/ansible/pull/80658/files#diff-9b710d1ef660617061e364a95a169b936dd95ed3c3710f2981412ed796133ae3R388

jpodivin:
Ansible documentation itself isn't compliant with that position. Adding description to endpoint may be nice, but many roles have only one of those, and their description would then be just modified description of the role itself. Effectively duplicating the effort and text for a little gain.

Furthermore. Even code curated directly by Ansible, including code used for tests, is not compliant with this approach. I haven't done a proper analysis. But a cursory github search suggests that it may be over half of all argument_spec.yml files out there [1]. 

With that in mind I don't believe the benefits of mandatory endpoint description, whatever they may be are justified.

[1]https://github.com/search?q=description%3A++path%3A**%2Fargument_specs.yml+owner%3Aansible&type=code&ref=advsearch

felixfontein:
> Adding description to endpoint may be nice, but many roles have only one of those, and their description would then be just modified description of the role itself.

Well, you can always provide `description: []` if you don't want to write one. But then you are explicitly saying "There is no description here". Or you can mention that users should look at the main entrypoint's description for information.

> Even code curated directly by Ansible, including code used for tests, is not compliant with this approach.

I don't see why the test files in ansible/ansible are relevant here. They are tests for ansible-core. They are not examples of how argument specs for roles should look like. ansible-core is very lenient in what it accepts, and that's fine - ansible-core is not a *documentation* tool, but an orchestration engine. But antsibull-docs is a *documentation* tool, so its assumption on role documentation can be a bit stricter.

PR #155, opened 2023-06-06
Semantic markup validator: also consider option aliases

felixfontein:
Right now something like
```yaml
   username:
     description:
     - Username that will be used to authenticate against InfluxDB server.
     - Alias O(login_username) added in Ansible 2.5.
     type: str
     default: root
     aliases: [ login_username ]
```
will cause validation to fail, since `login_username` is not considered a module/plugin option. (The validation in `ansible-test sanity --test validate-modules` already accepts this, so this PR makes the checks more similar.)


briantist (approved):

felixfontein:
@briantist thanks a lot for reviewing this!

PR #156, opened 2023-06-07
Fix deprecations for plugins and role entrypoints

felixfontein:
- Fix indent of 'Why' and 'Alternative' texts for deprecations. If these texts happen to be multiple lines long, the other lines must be indented.
- Allow role entrypoint deprecations without explicitly specifying `removed_from_collection` (which is the collection the role is part of anyway).

PR #157, opened 2023-06-08
Collection docs linting: validate references to other plugins/modules/roles

felixfontein:
Right now, `O()` and `RV()` are only validated for references to the same plugin/module/role. This PR extends this functionality to also validate references to other plugins/modules/roles in the same collection - including checking whether the names supplied to `M()` and `P()` make sense -, and optionally also allows to check references to dependent collections.

~~(I plan to extend this to check all installed collections as well, maybe already in this PR before it's ready, or in a follow-up PR...)~~

~~Currently includes #155 and #156, will rebase once these have been merged.~~


felixfontein:
This now contains all the features I want (for this PR :) ).

src/antsibull_docs/cli/antsibull_docs.py
```diff
@@ -575,6 +575,28 @@ def parse_args(program_name: str, args: list[str]) -> argparse.Namespace:
         " generation and schema and markup validation for"
         " plugins and roles in this collection.",
     )
+    lint_collection_docs_parser.add_argument(
+        "--validate-collection-refs",
+        dest="validate_collections_refs",
+        choices=["self", "dependent", "all"],
+        default="self",
```
felixfontein, line 582: I think I'm going to change the default to `dependent`, since that's (as opposed to `all`) pretty reproducible. What do folks think?

briantist: How are dependencies determined? Is that via `galaxy.yml`?

Part of me feels like it should be `all`; if I have references to collections I don't depend on, for whatever reason, I think I'd still like to know that those references are valid... but I'm having trouble understanding the danger or the downside to using that, could you explain a little more?

One more question: if the reference is to core does that get validated? with any option or only some of the options?

src/antsibull_docs/lint_plugin_docs.py
```diff
@@ -435,35 +567,78 @@ def __init__(
...
+    collection_to_plugin_info: Mapping[str, Mapping[str, Mapping[str, str]]],
+    collection_name: str,
+    dependencies: list[str],
+    collection_metadata: Mapping[str, AnsibleCollectionMetadata],
+    validate_collections_refs: ValidCollectionRefs,
+) -> _NameCollector:
+    name_collector = _NameCollector()
+    name_collector.collect_collection(collection_name)
+    for other_collection in collection_metadata.keys():
+        # ansible.builtin is always included in collection_metadata
```
briantist, line 602: I suppose this may have answered my above question about references to core?

src/antsibull_docs/cli/antsibull_docs.py
felixfontein: > How are dependencies determined? Is that via `galaxy.yml`?

Yes, and by looking at galaxy.yml of the dependent collection (the tool assumes they are all installed).

> Part of me feels like it should be `all`; if I have references to collections I don't depend on, for whatever reason, I think I'd still like to know that those references are valid... but I'm having trouble understanding the danger or the downside to using that, could you explain a little more?

The idea is that if you just run `antsibull-docs lint-collection-docs /path/to/collection/root --plugin-docs`, then the results are (mostly) independent on which collections you have exactly installed on your local machine.

Obviously the linting results depend on the exact versions of all dependent and transitively dependent collections, since options might not be there in older or - when removed - in newer versions. But if the collection refers to a module/option/... in an unrelated collection that is not a (transitive) dependency, and the default would be `all`, the linter could suddenly report errors that don't show up in CI - or the other way around, that another unrelated collection is installed in CI and there's an error due to that that won't show up locally when running `antsibull-docs lint-collection-docs`.

> One more question: if the reference is to core does that get validated? with any option or only some of the options?

You already figured out the "no". Maybe "dependent" should be changed to mean "dependent or ansible.builtin", since ansible.builtin is always around? I'm currently think extending it to include ansible.builtin is a good idea.

felixfontein: I have changed both (default and what `dependent` means) in 7c7166ba99fe2df983d20b427c3b0c3d1ff3b01d.

briantist (approved):

felixfontein:
@briantist thanks a lot!

PR #166, opened 2023-06-13
Improve handling of environment variables.

### LINKED ISSUE #164: How to handle environment variables?
With semantic markup, we have `E(...)` to reference environment variables. For RST output, the Sphinx `:envvar:` role is used (https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-envvar). antsibull-docs uses the `.. envvar::` directive (https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-envvar) for all environment variables that appear in `env:` plugin configs (and the Ansible/ansible-core docsite generates `.. envvar::` directives for all ansible-core config values, see https://github.com/ansible/ansible/blob/devel/docs/templates/config.rst.j2#L223-L248). But for all *other* environment variables mentioned in text, there is no such (auto-generated) target.

So either collections define their own targets, which can cause problems if multiple collections define targets that end up in the Ansible docsite, or they don't, and we end up with a list of `WARNING: 'envvar' reference target not found:` warnings. A good example is https://github.com/ansible-collections/community.docker/pull/645 with its output https://github.com/ansible-collections/community.docker/actions/runs/5235637304/jobs/9452611330.

This makes using `E(...)` inattractive or even outright impossible (for example in community.docker, the docsite build must have no errors for CI to pass). Manually defining `.. envvar::` directives is not a good idea either (except in special cases). #75 would improve the situation a bit, but it won't happen anytime soon (core doesn't like it).

I'm opening this issue so we can discuss what to do about this. Having a huge amount of new reference errors in the docsite build is something I'm pretty sure nobody wants. Just muting all these warnings (the one for env vars, not the general reference errors) isn't a good idea either IMO.

felixfontein:
Fixes #164.

This PR:
* Adds a new config field `envvar_directives` to the per-collection configuration. This allows the collection to declare which environment variables its extra docsite docs declare with a `.. envvar::` directive (example: https://github.com/ansible-collections/community.docker/pull/645/files#diff-2e6bfe479f7c2ad97ebb36910a39078e428365e0d6b29e7308ee3f255eda06e9 and https://github.com/ansible-collections/community.docker/pull/645/files#diff-51b1f7ed0708ccb6cda1e16d8b545eff7458d232c13fc144c47a64e9b404c39d).
* Adds two new roles `:ansenvvar:` (environment variable without reference) and `:ansenvvarref:` (environment variable with a reference; also is clever enough to only use the part before `=` to reference a `.. envvar::` directive).
* Modifies the RST generation to use these two new roles depending on whether an environment variable is deemed to be referencable or not. A variable is referencable if a) it shows up in a collection's `envvar_directives` config for one collection in the docsite; b) it is mentioned in the ansible-core config; c) it is mentioned in a plugin config in one of the collections used in the docsite.

Example: https://ansible.fontein.de/collections/community/docker/docker_container_module.html#notes - the environment variables mentioned in there are clickable and lead you to their definitions in https://ansible.fontein.de/collections/community/docker/docsite/scenario_guide.html#environment-variables

### DIFF
--- a/tests/functional/ansible-doc-cache-all-others.json
+++ b/tests/functional/ansible-doc-cache-all-others.json
@@ -1390,6 +1390,7 @@
         "Secret used to either login the ssh server or as a passphrase for ssh keys that require it",
         "Can be set from the CLI via the C(--ask-pass) option."
        ],
+       "type": "string",
        "vars": [
         {
          "name": "ansible_password"
@@ -1487,6 +1488,7 @@
          "version_added_collection": "ansible.builtin"
         }
        ],
+       "type": "string",
        "vars": [
         {
          "name": "ansible_private_key_file"
@@ -1518,6 +1520,7 @@
          "section": "paramiko_connection"
         }
        ],
+       "type": "string",
        "vars": [
         {
          "name": "ansible_paramiko_proxy_command",
@@ -1563,6 +1566,7 @@
        "description": [
         "Address of the remote target"
        ],
+       "type": "string",
        "vars": [
         {
          "name": "inventory_hostname"
@@ -1610,6 +1614,7 @@
          "name": "remote_user"
         }
        ],
+       "type": "string",
        "vars": [
         {
          "name": "ansible_user"
@@ -1642,6 +1647,7 @@
          "section": "ssh_connection"
         }
        ],
+       "type": "string",
        "vars": [
         {
          "name": "ansible_ssh_args",
@@ -1679,6 +1685,7 @@
          "version_added_collection": "ansible.builtin"
         }
        ],
+       "type": "string",
        "vars": [
         {
          "name": "ansible_ssh_common_args"
@@ -1714,6 +1721,7 @@
          "version_added_collection": "ansible.builtin"
         }
        ],
+       "type": "string",
        "vars": [
         {
          "name": "ansible_ssh_extra_args"
@@ -1987,7 +1995,7 @@
        ]
       },
       "ignore_proxy": {
-       "default": "no",
+       "default": false,
        "description": [
         "Will disable any environment proxy settings and connect directly to the remote host.",
         "This option is ignored if C(proxy) is set."
@@ -2341,6 +2349,7 @@
          "section": "ssh_connection"
         }
        ],
+       "type": "string",
        "vars": [
         {
          "name": "ansible_control_path",
@@ -2366,6 +2375,7 @@
          "section": "ssh_connection"
         }
        ],
+       "type": "string",
        "vars": [
         {
          "name": "ansible_control_path_dir",
@@ -2377,6 +2387,7 @@
       "host": {
        "default": "inventory_hostname",
        "description": "Hostname/IP to connect to.",
+       "type": "string",
        "vars": [
         {
          "name": "inventory_hostname"
@@ -2436,6 +2447,7 @@
       },
       "password": {
        "description": "Authentication password for the C(remote_user). Can be supplied as CLI option.",
+       "type": "string",
        "vars": [
         {
          "name": "ansible_password"
@@ -2504,6 +2516,7 @@
          "section": "ssh_connection"
         }
        ],
+       "type": "string",
        "vars": [
         {
          "name": "ansible_ssh_pkcs11_provider"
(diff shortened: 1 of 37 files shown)


tests/functional/ansible-doc-cache-all.json
```diff
@@ -9394,7 +9421,7 @@
...
...  ansible.builtin.apt_key:\n    url: https://ftp-master.debian.org/keys/archive-key-6.0.asc\n    state: present\n\n- name: Add an Apt signing key, will not download if present\n  ansible.builtin.apt_key:\n    id: 9FED2BCBDCD29CDF762678CBAED4B06F473041FA\n    url: https://ftp-master.debian.org/keys/archive-key-6.0.asc\n    state: present\n\n- name: Remove a Apt specific signing key, leading 0x is valid\n  ansible.builtin.apt_key:\n    id: 0x9FED2BCBDCD29CDF762678CBAED4B06F473041FA\n    state: absent\n\n# Use armored file since utf-8 string is expected. Must be of \"PGP PUBLIC KEY BLOCK\" type.\n- name: Add a key from a file on the Ansible server\n  ansible.builtin.apt_key:\n    data: \"{{ lookup('ansible.builtin.file', 'apt.asc') }}\"\n    state: present\n\n- name: Add an Apt signing key to a specific keyring file\n  ansible.builtin.apt_key:\n    id: 9FED2BCBDCD29CDF762678CBAED4B06F473041FA\n    url: https://ftp-master.debian.org/keys/archive-key-6.0.asc\n    keyring: /etc/apt/trusted.gpg.d/debian.gpg\n\n- name: Add Apt signing key on remote server to keyring\n  ansible.builtin.apt_key:\n    id: 9FED2BCBDCD29CDF762678CBAED4B06F473041FA\n    file: /tmp/apt.gpg\n    state: present\n",
```
briantist, line 9424: same here

tests/functional/ansible-doc-cache-all-others.json
```diff
@@ -9428,7 +9455,7 @@
...
...  ansible.builtin.apt_key:\n    url: https://ftp-master.debian.org/keys/archive-key-6.0.asc\n    state: present\n\n- name: Add an Apt signing key, will not download if present\n  ansible.builtin.apt_key:\n    id: 9FED2BCBDCD29CDF762678CBAED4B06F473041FA\n    url: https://ftp-master.debian.org/keys/archive-key-6.0.asc\n    state: present\n\n- name: Remove a Apt specific signing key, leading 0x is valid\n  ansible.builtin.apt_key:\n    id: 0x9FED2BCBDCD29CDF762678CBAED4B06F473041FA\n    state: absent\n\n# Use armored file since utf-8 string is expected. Must be of \"PGP PUBLIC KEY BLOCK\" type.\n- name: Add a key from a file on the Ansible server\n  ansible.builtin.apt_key:\n    data: \"{{ lookup('ansible.builtin.file', 'apt.asc') }}\"\n    state: present\n\n- name: Add an Apt signing key to a specific keyring file\n  ansible.builtin.apt_key:\n    id: 9FED2BCBDCD29CDF762678CBAED4B06F473041FA\n    url: https://ftp-master.debian.org/keys/archive-key-6.0.asc\n    keyring: /etc/apt/trusted.gpg.d/debian.gpg\n\n- name: Add Apt signing key on remote server to keyring\n  ansible.builtin.apt_key:\n    id: 9FED2BCBDCD29CDF762678CBAED4B06F473041FA\n    file: /tmp/apt.gpg\n    state: present\n",
```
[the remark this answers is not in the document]
felixfontein: It doesn't really matter, I just used a slightly different version of ansible-core devel to generate this dump than before.

I originally wanted to exclude the ansible.builtin stuff, but to properly test references I gave up on that idea and included them (that was part of #157).

tests/functional/ansible-doc-cache-all.json
felixfontein: Same thing here - a slightly different version of ansible-core. I'll resolve your two comments.

felixfontein:
@briantist thanks a lot for your review!

Issue #168, opened 2023-06-16
seealso plugin/module refs should be checked by linter


felixfontein:
Right now they aren't. The community.general.read_csv module was referencing a non-existing community.general.readcsv lookup (it should have been ansible.builtin.readcsv), and I only noticed due to Sphinx errors.

PR #169, opened 2023-06-20
Fix dict modifying error

felixfontein:
The construct
```
try:
  foo[bar]...
except KeyError:
  ...
```
is dangerous: if `foo` is a `defaultdict`, you end up with a new entry, instead of the `except` handler being executed. In case of docs augmentation, this can lead to a crash if a seealso without description is present for a non-existing plugin/module.


felixfontein:
This was crashing the tests while working on #168.

PR #173, opened 2023-06-20
Lint array stubs

felixfontein:
Semantic markup allows to specify array stubs `[]`, `[foo]`, etc. This extends the option/return value name linting so that
1. the stub syntax is checked better (the parser only removes all stubs and checks whether the remaining thing looks good, it thus also accepts `fo[]o` and treats it the same as `foo`);
2. it makes sure that `[]` is only used for lists and dictionaries (if `foo` is a list/dict and `bar` a suboption, accept `foo[].bar`), and in case of sub-options of lists enforces its usage (if `foo` is a list and `bar` a suboption, `foo.bar` is not OK, but `foo[].bar` is).


felixfontein:
One place this identified: https://github.com/ansible-collections/community.docker/pull/645/files/792280a7040e9c6ebc897308715ba06a33de7006..9b76a9e5c7d045e1e6e81458c85a7b16a519eb1e

PR #180, opened 2023-06-28
Add plugin reference role to the Sphinx extension and use it

felixfontein:
This makes it easier to reference to modules, plugins, and roles from extra docsite documentation. There's no longer a need to manually and explicitly compose references.


oraNod:
@felixfontein Hey I've been looking at this and it seems fine. TBH I'd like to understand things a little better, which might be easier over a chat than in a PR.

Anyway... having run tests with nox I'd like to see how this renders in generated HTML but not succeeding. Probably missing something obvious, got any tips? Thank you.

felixfontein:
@oraNod thanks for checking this out!

PR #194, opened 2023-08-05
Avoid permalink icon to be selectable

felixfontein:
Fixes part of https://github.com/ansible/ansible-documentation/issues/245.


samccann:
Okay I figured out how to test this locally, but now I don't know what page shows this problem? I thought it was the parameters table but those don't copy the link when I doubleclick on them  even w/o this patch...

felixfontein:
Hmm, actually this doesn't seem to be necessary at all. According to https://developer.mozilla.org/en-US/docs/Web/CSS/user-select#values, the default value is `auto`, and that means for `::after` that the text isn't selectable anyway.

The permalink icon reported in the original issue isn't in `::after`, but a regular DOM element.

I'll close this PR. Thanks everyone!

PR #199, opened 2023-08-19
Improve generated markup; add styling for LaTeX output

felixfontein:
~~Currently includes #195 and #198. Will rebase once these are merged.~~

Converts the 'inline' roles from the generated RST files to proper roles, moves the structural markup (bold, italics) from the stylesheet to the roles, and insert more raw LaTeX to use the same colors in the PDF than for things like type, required, version added, etc.

Screenshot from the [community.dns.hosttech_dns_zone_info module](https://docs.ansible.com/ansible/latest/collections/community/dns/hosttech_dns_zone_info_module.html#parameters):
image

PR #198, opened 2023-08-19
LaTeX builder: add indent for nested options and return values

felixfontein:
~~Currently includes #195 (modifies one of its changelog fragments). Will rebase once that is merged.~~

Uses `.. raw:: latex` to insert some hspace and a minipage when indentation of the left cell is needed. This isn't perfect, but a lot better than before (no indent at all).

Screenshot from the [community.dns.nameserver_record_info module return value docs](https://docs.ansible.com/ansible/latest/collections/community/dns/nameserver_record_info_module.html#return-values):
screenshot

PR #195, opened 2023-08-19
Improve Sphinx LaTeX builder support

felixfontein:
Ref: https://github.com/ansible/ansible-documentation/issues/109#issuecomment-1684914779

Relevant documentation: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-tabularcolumns - at the end of the section there's also a hint about `:class: longtable`.

PR #200, opened 2023-08-20
Convert collection links from raw HTML to directive

felixfontein:
Currently includes ~~#195,~~ #198, and #199. Will rebase once these are merged.

Converts the inline HTML for collection links to a custom directive. The external links are a custom nodes class inherited from docutils's `reference` with a custom HTML generator; for other builders, the `reference` generator provided by that builder is used (so in LaTeX it is a regular hyperlink reference). The "Communication" button though needs a Sphinx `pending_xref` node (to resolve the reference), which improves on the previous version by not relying on implementation details, but at the same time is worse than before because `pending_xref` is replaced by a plain `ref` (Sphinx doesn't allow to customize this), which rendered to HTML neither has `target="_blank"` nor `aria-role="button"`. (Aside from that, it looks and behaves like a button, though.)

The HTML output didn't change (except the missing attributes for the "Communication" button, which you cannot see here):
HTML output

The LaTeX output is now as expected - the first four buttons are hyperlinks, the last one an internal reference in the PDF file:
LaTeX output


felixfontein:
You can see the HTML result on:
- https://ansible.fontein.de/collections/community/general/index.html#description
- https://ansible.fontein.de/collections/community/general/aerospike_migrations_module.html#collection-links

changelogs/fragments/200-collection-links.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - "Create collection links with a custom directive. This makes them work with other builders than the HTML builder (https://github.com/ansible-community/antsibull-docs/pull/200)."
```
oraNod, line 2:
```suggestion
  - "Create collection links with a custom directive. This makes them compatible with builders other than the HTML builder (https://github.com/ansible-community/antsibull-docs/pull/200)."
```

src/sphinx_antsibull_ext/directives.py
```diff
@@ -33,8 +39,75 @@ def run(self):
...
+                f"Error while parsing content of {self.name} as YAML: {exc}"
+            ) from exc
+        if not isinstance(data, Sequence):
+            raise self.error(
+                f"Content of {self.name} must be a YAML list, got {data!r} - {content!r}"
+            )
+        for index, entry in enumerate(data):
+            if not isinstance(entry, Mapping):
+                raise self.error(
+                    f"Content of {self.name} must be a YAML list consisting of mappings:"
```
oraNod, line 75:
```suggestion
                    f"Content of {self.name} must be a YAML list of mappings:"
```

src/sphinx_antsibull_ext/directives.py
```diff
@@ -33,8 +39,75 @@ def run(self):
...
+            title = self._get_string(entry, index, "title", str)
+            url = self._get_string(entry, index, "url", str, optional=True)
+            ref = self._get_string(entry, index, "ref", str, optional=True)
+            external = (
+                self._get_string(entry, index, "external", bool, optional=True) or False
+            )
+            if (url is None) == (ref is None):
+                raise self.error(
+                    f"{index + 1}th entry in {self.name}:"
+                    " exactly one of 'url' and 'ref' must be provided"
```
oraNod, line 87:
```suggestion
                    " exactly one 'url' and one 'ref' must be provided"
```

oraNod (approved):
LGTM. Some really minor nits on wording. I couldn't find any faults or optimizations for the code side but I did look and it all seems fine. I think this is really nice work @felixfontein

src/sphinx_antsibull_ext/directives.py
felixfontein: This one should be an either-or: either 'url' is there, or 'ref' is there. Both aren't allowed, and not having any of them is also not allowed.

felixfontein: For me they mean exactly the same. Could you try to explain what's the difference for you?

oraNod: I can try my best to explain it. it's kind of something that just sounds "right" to me rather than it being a formal rule that I'm aware of.

basically "one of 'url' and 'ref'" conveys that you need one of each. not that you need either one url or one ref.

felixfontein: Ah, I see the problem. Would using `or` instead of `and` make it better? I'm wondering how to do this with three elements, like `Exactly one of a, b, or c must be provided`. Or can you simply write `Either a, b, or c must be provided`?

oraNod: I think the latter sounds clearer to me (`Either a, b, or c must be provided`). "One of" feels like something is missing in the construction, like `one of the following`.

felixfontein: Ah, that's much better! Thanks :)

PR #202, opened 2023-08-20
Update version and license information of Sphinx extension; make sure they are kept up to date

felixfontein:
Also store the version information in `src/antsibull_docs/__init__.py` so it can be used to embed it into the generated RST files, for example ("generated with antsibull-docs x.y.z"). I've been thinking of doing that for a longer time but wasn't able to get hold reliably of the version - using distutils or packaging doesn't work for local editable installs (or at least it didn't work when I tried it out a longer time ago, might have been related to poetry though).

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ nox
 ## Creating a new release:
 
 1. Run `nox -e bump -- <version> <release_summary_message>`. This:
-   * Bumps the package version in `pyproject.toml`.
+   * Bumps the package version in `pyproject.toml` and `src/sphinx_antsibull_ext/__init__.py`.
    * Creates `changelogs/fragments/<version>.yml` with a `release_summary` section.
    * Runs `antsibull-changelog release` and adds the changed files to git.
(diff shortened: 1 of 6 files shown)


felixfontein:
Unfortunately
```
import importlib.metadata

try:
    __version__ = importlib.metadata.version("antsibull_docs")
except importlib.metadata.PackageNotFoundError:
    __version__ = '0.dev0
```
is totally worthless for editable installs as I am using them. It reports the version that was in pyproject.toml when I last ran `pip install -e /path/to/antsibull_docs`.

Using `dynamic = ["version"]` would solve that problem, but then antsibull-changelog no longer can read the current version from `pyproject.toml`. Fortunately its `release` subcommand has a `--version` parameter.

gotmax23:
I would use `hatch version` to handle bumping the version. See https://paste.sr.ht/~gotmax23/80e9351701fdf6eabf2a24ea86f1870fe2b554d1.

gotmax23:
Retriggering CI. It seems something is wrong with coverage uploading.

gotmax23 (approved):

felixfontein:

PR #205, opened 2023-08-30
Add version into generated files

felixfontein:
The generated HTML by Sphinx contains something like `<meta content="2.3.1.post0" name="antsibull-docs">`.


src/antsibull_docs/test_helpers.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Determine whether the code is running in CI.
+    """
+    return (
+        os.environ.get(ANTSIBULL_DOCS_CI_ENV_MARKER_NAME)
+        == ANTSIBULL_DOCS_CI_ENV_MARKER_VALUE
+    )
+
+
+def ci_setup():
```
felixfontein, line 32: @gotmax23 I'm not really happy about this mechanism. I wanted to avoid some manual patching of `../../src/antsibull_docs/__init__.py` in `tests/functional/build-*.sh`. But maybe it's better to just do exactly that? What do you think?

gotmax23: You could create a shim Python script that imports `antsibull_docs`, monkeypatches `antsibull_docs.__version__`, and then calls `sys.exit(antsibull_docs.cli.antsibull_docs.main())`

felixfontein: Right. I wonder why I didn't think of that yesterday...

felixfontein: Ah wait, it wasn't yesterday, but today, early morning. Maybe that explains it ;)

felixfontein: Implemented in 39adce6eff12b8ebef8343c94aa332c2786d8681.

felixfontein:
This was discussed at today's DaWGs meeting, we agreed that the version should always be inserted for now. If someone comes up with a good reason why it should be possible to disable it, we can weight that reason against the additional complexity a switch needs.

PR #207, opened 2023-08-30
Add version subcommand

felixfontein:
Now that we have access to it, let's add the most obvious subcommand :)


gotmax23:
We already have a `--version` option provided by `antsibull_core.args.get_toplevel_parser()`. It uses the version from the dist-info metadata instead of `antsibull_docs.__version__`, though.

felixfontein:
Meh, I completely forgot about that one...

I guess I should work on a way to pass the real version into that one instead. Right now `antsibull-docs --version` for me shows 2.0.0a2, which was the current version when I last ran `pip install -e`...

felixfontein:
I created https://github.com/ansible-community/antsibull-core/pull/96 to be able to fix this.

PR #208, opened 2023-09-02
Update fixtures to current ansible-core docs; improve linting of array stubs

felixfontein:
This includes semantic markup for most plugin docs.

Once https://github.com/ansible/ansible/pull/81450 is merged and we can use that some of the unrelated information will vanish. (I'm really looking forward to that...)


felixfontein:
I had to add a code change since one of the current tests was now failing: `RV(ansible.builtin.stat#module:stat[foo.bar].exists)` is now correct, since the type of the ansible.builtin.stat module's return value `stat` changed from `complex` to `dict` - and dictionaries allow subscription with `[]`.

This shows another problem: while `stat` is a dictionary, `stat[something]` can be correct (depending on what exactly `something` is), while `stat[something].exists` is not correct (or at least not linkable to a documented return value). So I added a new rule to only allow array stubs for dictionaries if they show up as the last element (i.e. `foo.bar[baz]` is ok, but `foo[bar].baz` is not if `foo` is a dictionary).

felixfontein:
I'll merge in a couple of days if nobody objects by then.

gotmax23 (approved):

gotmax23:
Side note:

IMO, this code would be a easier to understand if

https://github.com/ansible-community/antsibull-docs/blob/cbb2c03e0ad83b3534eabfac67abab2d4f65a1a9/src/antsibull_docs/lint_plugin_docs.py#L379

was written along the lines of

``` python
    for index, (part, array_stub) in name:
        ...
```

Accessing the tuples returned by `split_option_like_name()` without unpacking them is a bit confusing, as the reader has to go look at that function's signature to figure out that it's a 2 tuple and what those two values refer to.

felixfontein:
Good point. I'll have to clean that up at some point.

felixfontein:
@gotmax23 thanks a lot for reviewing this!

PR #209, opened 2023-09-02
Bump dependency on antsibull-core and use new options

felixfontein:
Makes `antsibull-docs --help | grep version` and `antsibull-docs --version` work correctly, the latter also with long-lived editable installs.


gotmax23 (approved):

felixfontein:

PR #212, opened 2023-09-11
Restrict rstcheck to <6.2.0

devon-mar:
`rstcheck-core` requires `pydantic>=2` since v1.1.0 while antsibull-docs requires `pydantic<2`.

`rstcheck` v6.2.0 requires `rstcheck-core>=1.1` (https://github.com/rstcheck/rstcheck/commit/6834d0d708e5f305c13930d3eedeaab41057adb2).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
     "docutils",
     "jinja2 >= 3.0",
     "packaging",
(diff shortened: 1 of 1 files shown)


devon-mar:
Fixed by upgrading pip so that it resolves dependencies correctly.

Issue #211, opened 2023-09-09
Clarification on "expanded collections" and support for collections on disk


briantist:
When running `antsibull-docs collection --help` it says this about the collection list:

```
positional arguments:
  collections           One or more collections to document. If the names are directories on disk, they will be parsed as
                        expanded collections. Otherwise, if they could be collection names, they will be downloaded from galaxy.
```

I thought this might mean that you could give path names (where collections already exist) and that it would use those, retrieving the collection name from the `galaxy.yml` or `MANIFEST.json` in that path. But it does not seem to work that way, since it ends up trying to download the path as though it's a collection name.

> antsibull_core.galaxy.NoSuchCollection: No collection found at: https://galaxy.ansible.com/api/v2/collections/tmp/adocs/ansible_collections/namespace/collection/versions/

When I look through the code, it doesn't seem like there is any code supporting reading such a collection from disk, so I'm wondering if that note about the names being "directories on disk" refers to old functionality, or maybe it means something else entirely?

If I try to do the same with `--use-current`, the command "succeeds" but doesn't actually build anything except an empty `index.rst` and `environment_variables.rst`, presumably because it couldn't find collections with those weird names (there's no console output at all in this case, and `--fail-on-error` also didn't change that).

---

### Why I was interested in this

I'm going to build an internal docsite, and so I need some way to know which collections to include. 
Some will be installed via git rather than through the galaxy API protocol.

To organize this, I thought it might be nice to simply use a `requirements.yml` with the git sources to serve as my canonical collection list. If I install them in a separate step, then I can use `--use-current` to avoid auto-install, but then I need to supply the collection names themselves to `antsibull-docs` again.

I thought I could avoid it altogether with a workflow that looks something like:

```
ansible-galaxy collection install -r requirements.yml -p /tmp/adocs
antsibull-docs collection /tmp/adocs/ansible_collections/namespace/*
```

I wasn't too keen on parsing the YAML myself just to pass the names, especially since for git sources the `name:`  field is the git URL so I don't necessarily have a name to pass without storing it separately.

### Suggestions

For now I will just have to keep a separate list of collection names probably, but I can imagine two possible enhancements:
1. You're already attempting to install collections from the list of names, maybe have a mode where it takes a `requirements.yml` file instead and can install using that config, and then no names need to be passed.
2. Implement/restore the ability to pass directories on disk that already refer to collections.

If you think either of those are a possibility I can split them out to separate issues, curious what you think.

[...]

PR #215, opened 2023-09-15
Allow to exclude ansible.builtin from 'current' subcommand

felixfontein:
Ref: #211.

### DIFF
--- a/changelogs/fragments/215-current-skip-ansible-builtin.yml
+++ b/changelogs/fragments/215-current-skip-ansible-builtin.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "The ``current`` subcommand now has a ``--skip-ansible-builtin`` option which skips building documentation for ``ansible.builtin`` (https://github.com/ansible-community/antsibull-docs/pull/215)."
--- a/src/antsibull_docs/cli/antsibull_docs.py
+++ b/src/antsibull_docs/cli/antsibull_docs.py
@@ -399,6 +399,12 @@ def parse_args(program_name: str, args: list[str]) -> argparse.Namespace:
         " specified, all collections in the currently configured ansible"
         " search paths will be used",
     )
+    current_parser.add_argument(
+        "--skip-ansible-builtin",
+        action="store_true",
+        help="Do not generate documentation for ansible.builtin, the synthetic"
+        " collection consisting of all modules and plugins bundled with ansible-core.",
+    )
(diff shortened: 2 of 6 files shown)


briantist (approved):
This looks great to me! Only question I have is about whether it makes sense to allow excluding any number of arbitrary collections. It seems like the internal function is already changed for that, it would be a matter of exposing it with like `--excluded-collections` or something.

`--skip-ansible-builtin` is still a nice shorthand option to specifically skip that; I could understand not wanting to expose a wider exclude right now due to complexity/testing, just wondering if there was any other reason.

felixfontein:
> Only question I have is about whether it makes sense to allow excluding any number of arbitrary collections. It seems like the internal function is already changed for that, it would be a matter of exposing it with like `--excluded-collections` or something.

I thought about exposing that, but I'm not sure it's a great interface. While it works OK for some specific situations, it's generally not a useful interface since more often you want to restrict to a specific set of collections, instead of excluding a specific set. The set of dependencies of a specific set of collections can change all the time, so if you exclude all dependencies there are *right now* the docs build in a month might suddenly include something you don't want it to include.

I think it's better to first try to figure out a better interface, and only allow this basic `--exlcluded-collections` if we cannot find a better way to do this.

felixfontein:

briantist:
That's interesting, because you were thinking about excluding mostly for dependencies, and I wasn't even thinking about that (though it makes a lot of sense), I was thinking mostly about quickly building docs from and existing set of collections but also wanting to exclude certain ones even if they weren't dependencies... but I guess it's not going to be a common usage.

But I agree working out the interface(s) makes sense.

In a way, namespaces cold be more useful for building certain types of docsites. For example the one I'm building is internal and every collection is expected to use the same namespace. If collection names (for inclusion or exclusion) were wildcards, similar to how it is in the `antsibull-docs.cfg` file, that could be pretty versatile.

If I could do `antsibull-docs collection --use-current namespace.* other_ns.collection` or `antsibull-docs current --include-collections namespace.*` (maybe `--filter-collections`?) that would be cool.

felixfontein:
Wildcard support for `antsibull-docs collection --use-current` sounds like a good idea. It's also something that can be implemented - first `ansible-galaxy collection list` can be used to get a list of all available collections (which is done anyway to figure out where the collections are located), and that can be used to figure out the actual list of collections.

PR #216, opened 2023-09-17
Improve logical information for labels to Ansible objects

### LINKED ISSUE #203: Improve logical information in Sphinx indexes
I tried creating a docset for a collection for use in Dash or Zeal with [doc2dash](https://doc2dash.hynek.me/en/stable/), and noticed that while the process works pretty well (except that the default theme isn't really suitable for this; according to https://gist.github.com/danieltomasz/d6aae37a0ba5f5ba0f17ea68b590731f the epub theme is better suited), with one big exception - the 'Sections' list:
image

The `n` copies of "Requirements" isn't exactly helpful (these are the requirements for every plugin and module), and the raw (internal) labels for plugins/modules/roles and their parameters and return values are even less helpful.

The latter can be improved by not adding labels to texts, but instead using a custom directive which generates the label and adds a nice title to it.

These things could also be added to an index (if you generate a PDF, you see that right now only the environmen variables show up in the index, and it is otherwise empty).

felixfontein:
Fixes #203.

Current version:
image

The n-fold 'Requirements' section still shows up, but at least the references to all return values and options look **much** better! Also, there's now an index with all modules, plugins, and roles (if you tell Sphinx to generate it):

image


felixfontein:
More Zeal screenshots:

The requirements references look better:
image

Return values for function-like plugins (lookups, filters, tests):
image

Nested return values for function-like plugins (lookups, filters, tests):
image

Building the official docsite with an appropriate theme and converting it with doc2dash should now produce a pretty neat docset.

felixfontein:
One problem is that the current solution works well with HTML generation, but not at all with LaTeX. I tried to play around with that the last days, but didn't had much time, and the only way I got it to work with both HTML and LaTeX writers resulted in a large amount of WARNINGS and CRITICAL errors (which did not impact functionality, since the collisions they complained about were intentional).

I'll try to spend some more time figuring out how to do this properly...

felixfontein:
I think a better way is to allow doc2dash to change/remove labels with a (say) JSON file, and then dump such a JSON file from antsibull-docs. That allows us to obtain all these extra details *without* doing dirty tricks to try to make Sphinx output the titles for references that we want in the Interspinx file.

PR #217, opened 2023-09-17
Deprecate --use-html-blobs

felixfontein:
This feature is causing more and more problems, in particular now that it's so easy to reference options and return values (which simply won't work when this feature is enabled).

It is possible to make it work, but it's quite some amount of work, and I'm convinced that this isn't really needed.

(HTML tables are still a thing for the `simplified-rst` output format and won't go away there.)

### DIFF
--- a/changelogs/fragments/217-deprecate-use-html-blobs.yml
+++ b/changelogs/fragments/217-deprecate-use-html-blobs.yml
@@ -5,4 +5,4 @@ deprecated_features:
      values. If you think this feature needs to stay, please create an issue in the `antsibull-docs
      repository <https://github.com/ansible-community/antsibull-docs/issues/>`__ and provide
      good reasons for it
-     (https://github.com/ansible-community/antsibull-docs/pull/216)."
+     (https://github.com/ansible-community/antsibull-docs/pull/217)."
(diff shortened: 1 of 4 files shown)


changelogs/fragments/217-deprecate-use-html-blobs.yml
```diff
@@ -0,0 +1,8 @@
+deprecated_features:
+  - "The ``--use-html-blobs`` feature that inserts HTML blobs for the options and return value
+     tables for the ``ansible-docsite`` output format is deprecated and will be removed soon.
+     The HTML tables cause several features to break, such as references to options and return
+     values. If you think this feature needs to stay, please create an issue in the `antsibull-docs
+     repository <https://github.com/ansible-community/antsibull-docs/issues/>`__ and provide
+     good reasons for it
+     (https://github.com/ansible-community/antsibull-docs/pull/216)."
```
briantist, line 8:
```suggestion
     (https://github.com/ansible-community/antsibull-docs/pull/217)."
```

gotmax23:
I would print these warnings to `file=sys.stderr`

briantist:
that is a good point, why not use `warnings.warn`?

felixfontein:
> that is a good point, why not use `warnings.warn`?

I just tried that, the result was that the warning isn't printed.

briantist (approved):

felixfontein:
(I didn't spend much time trying to figure out *why* the warnings don't show... But I guess doing it this way also works, since these parts aren't used - or shouldn't be used - as Python libraries.)

felixfontein:
@briantist @gotmax23 thanks a lot for reviewing this!

PR #219, opened 2023-09-25
Allow wildcards for collections subcommand when --use-current is used

felixfontein:
Ref: https://github.com/ansible-community/antsibull-docs/pull/215#issuecomment-1722474667


briantist (approved):
This is great! I would definitely use this.

felixfontein:

briantist:
Thanks @felixfontein , do you have an idea when this might be released? A pre-release is also something I might be able to start using.

PR #224, opened 2023-12-04
add compatibility for pydantic v2

gotmax23:
This uses the `pydantic.v1` compat module (well really, it's just an
in-tree copy of the latest pydantic v1 release) so we can also support
pydantic v2. Major changes to root models make it infeasible to support
pydantic v2 natively without dropping support for v1.

The `_pydantic_compat` module simplifies imports of the `pydantic.v1`
compat module with pydantic v2. The pydantic.v1 compat module is always
used when type checking. We have to force one of them when type
checking, as type checkers cannot understand `try-except` imports.

---

See https://github.com/ansible-community/antsibull-core/issues/125.

### DIFF
--- a/.pylintrc.automated
+++ b/.pylintrc.automated
@@ -449,7 +449,7 @@ max-branches=12
 max-locals=15
 
 # Maximum number of parents for a class (see R0901).
-max-parents=7
+max-parents=8
 
 # Maximum number of public methods for a class (see R0904).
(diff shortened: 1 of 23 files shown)


gotmax23:
Ready for review

felixfontein (approved):

felixfontein:
@gotmax23 thanks for this one as well!

PR #225, opened 2023-12-08
Detect the EXAMPLES format and use it

felixfontein:
Currently the ansible-documentation build fails since https://github.com/ansible/ansible/pull/82361 got merged. This PR fixes the handling of `EXAMPLES` blocks by using the format specification if provided.

Ref: https://github.com/ansible/ansible/pull/82355#issuecomment-1841469933
Ref: https://github.com/ansible/ansible/blob/6ebefaceb6cd0d4961776a94d63a71fc1fc28bc0/test/lib/ansible_test/_util/controller/sanity/yamllint/yamllinter.py#L173-L187


samccann (approved):
Can't speak to the code but compared before/after examples and LGTM thanks!

src/antsibull_docs/schemas/docs/plugin.py
```diff
@@ -207,6 +210,16 @@ class PluginDocSchema(BaseModel):
...
+    examples_format: str = "yaml"
+
+    @p.root_validator(pre=True)
+    # pylint:disable=no-self-argument
+    def extract_examples_format(cls, values):
+        if isinstance(values.get("examples"), str):
+            fmt_match = _EXAMPLES_FMT_RE.match(values["examples"].lstrip())
+            if fmt_match:
+                values["examples_format"] = fmt_match.group(1)
+        return values
```
gotmax23, line 222: I would reweite this as a regular field validator and maybe use the walrus syntax for the regex match if that's your thing:

```suggestion
    @p.validator(pre=True, always=True)
    # pylint:disable=no-self-argument
    def extract_examples_format(cls, value: t.Any, values: dict[str, Any]):
        if isinstance(value, str) and fmt_match := _EXAMPLES_FMT_RE.match(values["examples"].lstrip()):
            value = fmt_match.group(1)
        return value
```

felixfontein: I replaced it, though I had to modify your code a bit. `:=` does not work in `if X and Y := Z:`, that results in a syntax error.

gotmax23: If you use `if isinstance(...) and (fmt_match := _EXAMPLES_FMT_RE.match(values["examples"].lstrip())` with parentheses around the second expression, it should work.

I'm not a huge fan of


``` python
if CONIDTION:
   if CONDITION2:
       ...
...
```

Other than that, this looks good to me.

felixfontein: Due to the line length limit this gets reformatted as:
```py
        if isinstance(examples := values.get("examples"), str) and (
            fmt_match := _EXAMPLES_FMT_RE.match(examples.lstrip())
        ):
            value = fmt_match.group(1)
```
Which IMO is less readable than two nested `if`s.

src/antsibull_docs/schemas/docs/plugin.py
```diff
@@ -207,6 +210,15 @@ class PluginDocSchema(BaseModel):
...
     examples: str = ""
+    examples_format: str = "yaml"
+
+    @p.validator("examples_format", always=True)
+    # pylint:disable=no-self-argument
+    def extract_examples_format(cls, value: t.Any, values: dict[str, t.Any]):
+        if isinstance(examples := values.get("examples"), str):
+            if fmt_match := _EXAMPLES_FMT_RE.match(examples.lstrip()):
+                value = fmt_match.group(1)
+        return value
```
gotmax23, line 221: > Due to the line length limit this gets reformatted as [...] which IMO is less readable than two nested ifs.

Yeah, that's ugly. Black will leave this alone, though:

```suggestion
        # fmt: off
        if (
            isinstance(examples := values.get("examples"), str) and
            (fmt_match := _EXAMPLES_FMT_RE.match(examples.lstrip()))
        ):
        # fmt: on
            value = fmt_match.group(1)
        return value
```

felixfontein: I dislike formatter directives even more than all the other solutions :) Do you mind just keeping the two `if`s?

gotmax23: Alright, sure.

gotmax23 (approved):

gotmax23:
Thanks!

felixfontein:
@gotmax23 thanks for reviewing, merging, and releasing!

PR #229, opened 2023-12-16
Use correct length also for wide Unicode letters for title underlines

### LINKED ISSUE #228: The multibyte role `short_description` causes warning "Title underline too short."
The multibyte role `short_description` causes warning "Title underline too short."

`argument_specs.yml`: 

```yaml
---
argument_specs:
  main:
    short_description: short description of Main entry point for テストロール
    description:
      - description of Main entry point for テストロール
    author:
      - akira6592
    options:
      myapp_int:
        type: "int"
        required: false
        default: 42
        description:
          - "The integer value, defaulting to 42."
```

```sh
% dest/build.sh

...(omit)...
/Users/akira/ansible/dest/rst/collections/local/mycollection/myapp_role.rst:19: WARNING: Title underline too short.

local.mycollection.myapp role -- short description of Main entry point for テストロール
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
/Users/akira/ansible/dest/rst/collections/local/mycollection/myapp_role.rst:41: WARNING: Title underline too short.

Entry point ``main`` -- short description of Main entry point for テストロール
------------------------------------------------------------------------

...(omit)...
```

It is not only role `short_description`, but maybe also other `short_description`.

felixfontein:
Fixes #228.

### DIFF
--- a/src/antsibull_docs/jinja2/filters.py
+++ b/src/antsibull_docs/jinja2/filters.py
@@ -14,6 +14,7 @@ import typing as t
 from collections.abc import Mapping, Sequence
 
 from antsibull_core.logging import log
+from docutils.utils import column_width as _column_width  # pyre-ignore[21]
 from jinja2.runtime import Context, Undefined
 from jinja2.utils import pass_context
 
@@ -251,3 +252,7 @@ def rst_format(fmt: str, for_sphinx: bool = False) -> str:
         return "json"
     # Unknown format: no syntax highlighting
     return "text"
+
+
+def column_width(string: str) -> int:
+    return _column_width(string)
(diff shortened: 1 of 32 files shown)


src/antsibull_docs/jinja2/filters.py
```diff
@@ -14,6 +14,7 @@
 from collections.abc import Mapping, Sequence
 
 from antsibull_core.logging import log
+from docutils.utils import column_width as _column_width  # pyre-ignore[21]
```
gotmax23, line 17: Why not just import this as `column_width` and remove the `def column_width` that just returns the result of `docutils.utils.column_width`?

felixfontein: Just importing it without using it in the module makes the linters complain that this is an unused import. This way it's clear that it is used.

gotmax23: Adding it to `__all__` would fix that problem, but we don't already have that set for this module, so I think this is fine. Thanks for the explanation!

/resolve

gotmax23 (approved):

felixfontein:

PR #231, opened 2023-12-19
add pre-commit.com hooks for lint-collection-docs

gotmax23:


felixfontein (commented):
I'm not sure this is a good idea.
1) Who installs ansible-core?
2) Who installs the required collections (for docs fragments, references, ...)?

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,15 @@
...
+- id: lint-collection-docs
+  name: Lint collection docs
+  language: python
+  pass_filenames: false
+  entry: "antsibull-docs lint-collection-docs ."
+  # Default arguments based on the collection_template.
+  # Users can change these if they wish
+  args:
+    - "--plugin-docs"
+    - "--skip-rstcheck"
```
felixfontein, line 15: This should make sure ansible-core is installed:
```suggestion
    - "--skip-rstcheck"
  additional_dependencies:
    - ansible-core >= 2.14.0
```

ssbarnea:
> I'm not sure this is a good idea.
> 
> 1. Who installs ansible-core?
> 2. Who installs the required collections (for docs fragments, references, ...)?

I like the idea but there are some problems with using with pre-commit:

- pre-commit.ci does not allow network access, so installing a collection would not be possible with them. Still, this does not mean we cannot use pre-commit itself and run it in our own CI/CD, is just that we need to skip running it on that SaaS.

- By definition running pre-commit should not ask user to to other preparatory tasks, like installing collection. This tells me that we need to do it inside the hook, so it would self-contained.

felixfontein (changes requested):
Leaving a negative review so this won't get merged prematurely.

PR #233, opened 2023-12-23
Use certain fields from library context instead of app context that are deprecated in the app context and are removed from antsibull-core 3.0.0

felixfontein:
These fields have been moved from app to library context for antsibull-core 2.0.0 (or even before?). I apparently forgot to fix this earlier...

Ref: https://github.com/ansible-community/antsibull-core/pull/128


gotmax23 (approved):

felixfontein:
@gotmax23 thanks for reviewing this one as well!

PR #242, opened 2024-01-30
Add new plugin_name to DocSchema

sivel:
In https://github.com/ansible/ansible/commit/bec27fb4c0a40c5f8bbcf26a475704227d65ee73 we added a new `plugin_name` field to the `ansible-doc` output. This PR adds this new field to the `DocSchema` class

See https://github.com/ansible/ansible-documentation/issues/1079

### DIFF
--- a/src/antsibull_docs/schemas/docs/base.py
+++ b/src/antsibull_docs/schemas/docs/base.py
@@ -619,6 +619,7 @@ class DocSchema(BaseModel):
     collection: str = REQUIRED_COLLECTION_NAME_OR_EMPTY_STR_F
     description: list[str]
     name: str
+    plugin_name: str
     short_description: str
     aliases: list[str] = []
     author: list[str] = []
--- a/src/antsibull_docs/schemas/docs/base.py
+++ b/src/antsibull_docs/schemas/docs/base.py
@@ -619,7 +619,7 @@ class DocSchema(BaseModel):
     collection: str = REQUIRED_COLLECTION_NAME_OR_EMPTY_STR_F
     description: list[str]
     name: str
-    plugin_name: str
+    plugin_name: str = ""
     short_description: str
     aliases: list[str] = []
     author: list[str] = []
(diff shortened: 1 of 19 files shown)


gotmax23:
@felixfontein, we should also consider getting rid of the `extra="forbid"` config. I don't think we need to be that strict here.

src/antsibull_docs/schemas/docs/base.py
```diff
@@ -619,6 +619,7 @@ class DocSchema(BaseModel):
     collection: str = REQUIRED_COLLECTION_NAME_OR_EMPTY_STR_F
     description: list[str]
     name: str
+    plugin_name: str = ""
```
gotmax23, line 622: I think it'd make more sense to set this to `None` if it's not part of the docs schema.

```suggestion
    plugin_name: t.Optional[str] = None
```

sivel: `filename` is also the same, it's added by ansible-doc and not part of the docs schema itself, was following that as my example.

gotmax23:
> This seems to be indicating a change to pydantic, and nothing I've done.

There was a new pydantic release yesterday that's likely the culprit. pydantic does some funky module-level `__getattr__()` stuff that sometimes confuses static analysis tools. I'm not sure why pylint thinks a module is a function...

sivel:
I've pushed a fix for the pylint issue in this PR.  Except for now some codecov stuff...

src/antsibull_docs/schemas/docs/base.py
gotmax23: Okay, fair enough.

gotmax23 (approved):

felixfontein:
> @felixfontein, we should also consider getting rid of the `extra="forbid"` config. I don't think we need to be that strict here.

For the docs rendering part, ok. For the linting part, we must be strict.

felixfontein:
Looks OK, except that it's missing a changelog fragment. I can push one.

felixfontein (approved):

felixfontein:
@sivel thanks for fixing this!

gotmax23:
Thanks everyone!

PR #253, opened 2024-03-06
Added Dark Mode support to option tables

SebastianGode:
I noticed that Antsibull didn't respect the `prefers-color-scheme` browser or system setting of the users which access the documentation website.
This will result in completely unreadable tables for Sphinx-Themes which already include support for such user settings.

To fix this I added some css code to change the colors accordingly for dark mode. The colors are chosen to get a contrast ration >7 and to pass Google's usability rating with an AAA.

If you have any suggestions feel free to mention them.

Before
image
After
image

Tip: You can quickly switch the Browser View to Dark-Mode or Light-Mode by using the Developer Options:
image
Just don't use the auomatic conversion from the Browser as that will replace all CSS.


felixfontein:
@SebastianGode thanks for your contribution! I'll take a closer look soon. Can you add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

changelogs/fragments/253-darkmode-support-to-option-tables.yaml
```diff
@@ -0,0 +1,3 @@
+---
+minor_changes:
+  - Add support for "Dark-Mode" inside the Option-Tables (https://github.com/ansible-community/antsibull-docs/pull/253).
```
felixfontein, line 3:
```suggestion
  - Add support for "dark mode" to the option table styling (https://github.com/ansible-community/antsibull-docs/pull/253).
```

felixfontein (commented):
There are two big problems with this PR:

1. This is only setting *some* colors. In case the theme itself does not have a dark mode, the tables are partially unreadable: 
image
2. In low-width modes, it really looks bad since most colors are not adjusted: 
image

Maybe it's better to make dark mode support configurable, so you can enable it if the underlying theme also supports it, and disable it if not. That would fix 1. For 2., some more CSS would be needed.

felixfontein:
I started working around these problems in #254. That PR makes it easy to add new color schemes (this PR would translate into creating a copy of the default scheme, adding a media query with redefinitions of the colors for dark mode, and allowing to select that color scheme in the config).

That way colors can be selected (or even defined) by the extension's user (who also has control over which theme and other extensions are used).

felixfontein:
I went ahead and merged #254 so you aren't blocked on that PR. Note that I changed the way the node.js dependencies (sass and postcss) are installed, now you have to run `npm install` or `npm clean-install` in src/sphinx_antsibull_ext/css/ before running `./build.sh`.

SebastianGode:
@felixfontein I now think that I have fixed the tables in the narrow view. And adding the option to select different themes was awesome! That works like magic now :)

src/sphinx_antsibull_ext/assets.py
```diff
@@ -57,6 +57,7 @@
 _AVAILABLE_COLOR_SCHEMES = [
     "none",
     "default",
+    "default-autodark"
```
felixfontein, line 60:
```suggestion
    "default-autodark",
```

felixfontein:
@SebastianGode thanks a lot for your contribution!

SebastianGode:
@felixfontein Would it be possible to get a new Release to PyPi?  😃

felixfontein:
Yes, maybe later today or tomorrow. I want to add a small change before releasing.

PR #256, opened 2024-03-10
CI: nox: set up environments in a separate step

felixfontein:
This should clean up the actual test output without having to remove `-v`.


felixfontein:
@gotmax23 what do you think of this solution?

gotmax23 (approved):
LGTM. Thanks for working on this!

In a follow up PR, we should probably separate the two jobs into a matrix like we have in the antsibull repository to avoid code duplication.

felixfontein:
Thanks for reviewing this!

felixfontein:
Adding the matrix in #259.

PR #258, opened 2024-03-12
Refactor color schemes to avoid repetition; add dark color scheme

felixfontein:
Extends #253.

Ref: https://sass-lang.com/documentation/at-rules/use/


felixfontein:
CC @SebastianGode what do you think?

SebastianGode:
@felixfontein This is awesome!
I might start to like scss, reusing color variables is really cool.

felixfontein:
@SebastianGode thanks for taking a look!

PR #267, opened 2024-03-31
Allow collections to render changelog as part of their docsite

### LINKED ISSUE #31: Render changelogs in docsite build

felixfontein:
Fixes #31.

To do:
- [x] Add documentation.
- [x] Add proper error handling.
- [x] Make configuration more future-proof.
- [x] Discuss at DaWG meeting where to place changelog on collection index page.
- [x] Add tests.

### DIFF
--- a/src/antsibull_docs/schemas/collection_config.py
+++ b/src/antsibull_docs/schemas/collection_config.py
@@ -20,3 +20,6 @@ class CollectionConfig(p.BaseModel):
     # List of environment variables that are defined by `.. envvar::` directives
     # in the extra docsite RST files.
     envvar_directives: list[str] = []
+
+    # Whether to include the collection's changelog (added in version 2.10.0)
+    include_changelog: bool = False
--- a/src/antsibull_docs/schemas/collection_config.py
+++ b/src/antsibull_docs/schemas/collection_config.py
@@ -12,6 +12,11 @@
 from antsibull_docs._pydantic_compat import v1 as p
 
 
+class ChangelogConfig(p.BaseModel):
+    # Whether to write the changelog
+    write_changelog: bool = False
+
+
 class CollectionConfig(p.BaseModel):
     # Whether the collection uses flatmapping to flatten subdirectories in
     # `plugins/*/`.
@@ -21,5 +26,5 @@ class CollectionConfig(p.BaseModel):
     # in the extra docsite RST files.
     envvar_directives: list[str] = []
 
-    # Whether to include the collection's changelog (added in version 2.10.0)
-    include_changelog: bool = False
+    # Changelog configuration (added in version 2.10.0)
+    changelog: ChangelogConfig = ChangelogConfig()
--- a/tests/functional/baseline-no-breadcrumbs/collections/ns/col2/index.rst
+++ b/tests/functional/baseline-no-breadcrumbs/collections/ns/col2/index.rst
@@ -35,6 +35,7 @@ Description
     :maxdepth: 1
 
 
+
 Plugin Index
 ------------
 
--- a/tests/functional/baseline-no-breadcrumbs/collections/ns/col2/index.rst
+++ b/tests/functional/baseline-no-breadcrumbs/collections/ns/col2/index.rst
@@ -34,8 +34,6 @@ Description
 .. toctree::
     :maxdepth: 1
 
-
-
 Plugin Index
 ------------
 
--- a/tests/functional/baseline-no-breadcrumbs/collections/ns2/col/changelog.rst
+++ b/tests/functional/baseline-no-breadcrumbs/collections/ns2/col/changelog.rst
@@ -1,0 +1,4 @@
+
+The changelog of ns2.col could not be rendered::
+
+  list index out of range
--- a/tests/functional/baseline-no-breadcrumbs/collections/ns2/flatcol/index.rst
+++ b/tests/functional/baseline-no-breadcrumbs/collections/ns2/flatcol/index.rst
@@ -30,8 +30,6 @@ Description
 .. toctree::
     :maxdepth: 1
 
-
-
 Plugin Index
 ------------
 
--- a/tests/functional/baseline-no-breadcrumbs/collections/ns2/flatcol/index.rst
+++ b/tests/functional/baseline-no-breadcrumbs/collections/ns2/flatcol/index.rst
@@ -31,6 +31,7 @@ Description
     :maxdepth: 1
 
 
+
 Plugin Index
 ------------
 
--- a/tests/functional/baseline-no-indexes/collections/ns2/flatcol/index.rst
+++ b/tests/functional/baseline-no-indexes/collections/ns2/flatcol/index.rst
@@ -30,6 +30,7 @@ Description
     :maxdepth: 1
 
 
+
 Plugin Index
 ------------
 
--- a/tests/functional/baseline-no-indexes/collections/ns2/flatcol/index.rst
+++ b/tests/functional/baseline-no-indexes/collections/ns2/flatcol/index.rst
@@ -29,8 +29,6 @@ Description
 .. toctree::
     :maxdepth: 1
 
-
-
 Plugin Index
 ------------
 
--- a/tests/functional/baseline-simplified-rst/collections/ns/col2/index.rst
+++ b/tests/functional/baseline-simplified-rst/collections/ns/col2/index.rst
@@ -25,8 +25,6 @@ Description
 
 
 
-
-
 Plugin Index
 ------------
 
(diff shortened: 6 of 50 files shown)


felixfontein:
You can see the result here: https://ansible-collections.github.io/community.dns/pr/192/index.html#changelog (belongs to https://github.com/ansible-collections/community.dns/pull/192).

felixfontein:
This is now ready for review! (I will only mark the PR as that on GitHub once the DaWGs meeting was able to talk about this.)

tests/functional/baseline-no-breadcrumbs/collections/ns/col2/index.rst
```diff
@@ -35,6 +35,7 @@ Description
     :maxdepth: 1
 
 
+
```
briantist, line 38: what's this extra line for?

tests/functional/baseline-no-breadcrumbs/collections/ns2/col/changelog.rst
```diff
@@ -0,0 +1,4 @@
+
+The changelog of ns2.col could not be rendered::
+
+  list index out of range
```
briantist, line 4: is this intentional? part of a test case?

tests/functional/baseline-no-breadcrumbs/collections/ns2/flatcol/index.rst
```diff
@@ -31,6 +31,7 @@ Description
     :maxdepth: 1
 
 
+
```
briantist, line 34: extra line?

tests/functional/baseline-no-indexes/collections/ns2/flatcol/index.rst
```diff
@@ -30,6 +30,7 @@ Description
     :maxdepth: 1
 
 
+
```
briantist, line 33: some linter/formatter in your editor?

tests/functional/baseline-simplified-rst/collections/ns/col2/index.rst
```diff
@@ -26,6 +26,7 @@ Description
 
 
 
+
```
briantist, line 29: the mystery deepens....

tests/functional/baseline-simplified-rst/collections/ns2/col/changelog.rst
```diff
@@ -0,0 +1,4 @@
+
+The changelog of ns2.col could not be rendered::
+
+  list index out of range
```
briantist, line 4: also here (wondering if intentional)

tests/functional/baseline-simplified-rst/collections/ns2/flatcol/index.rst
```diff
@@ -22,6 +22,7 @@ Description
 
 
 
+
```
briantist, line 25: 👀

tests/functional/baseline-use-html-blobs/collections/ns2/col/changelog.rst
```diff
@@ -0,0 +1,4 @@
+
+The changelog of ns2.col could not be rendered::
+
+  list index out of range
```
briantist, line 4: ?

src/antsibull_docs/schemas/collection_config.py
```diff
@@ -12,6 +12,11 @@
 from antsibull_docs._pydantic_compat import v1 as p
 
 
+class ChangelogConfig(p.BaseModel):
+    # Whether to write the changelog
+    write_changelog: bool = False
```
briantist, line 17: Do we think we might ever default this to true? maybe in a new major version? 
I guess we'd want to discuss in the forum and/or dawgs meeting.

If so, we could announce the future default value in the changelog in this PR too

briantist (approved):
mostly looks good to me! been looking forward to this for a long time, I'll be able to remove my symlinks and extra docs entries in a bunch of collections, thanks for implementing!

tests/functional/baseline-no-breadcrumbs/collections/ns2/col/changelog.rst
felixfontein: This is intentional, for this collection the changelog is broken (changelogs/changelog.yaml contains invalid data which triggers a crash).

src/antsibull_docs/schemas/collection_config.py
felixfontein: It should never default to true, but it could default to something like "auto" which checks whether changelogs/changelog.yaml is present and only then sets it to true. But that would have to wait for a new major release.

Also I don't think the default should switch; I'd rather have a configurable toggle for the docsite build which allows to override this for collections.

tests/functional/baseline-no-breadcrumbs/collections/ns/col2/index.rst
felixfontein: This comes from the empty line between:
```jinja2
{% endif %}

{% for section in extra_docs_sections %}
```

felixfontein: Removed in cb2c5873b542650208a79aa19922346d40ae6ad6.

briantist (approved):

oraNod (approved):
Some minor comments. @felixfontein my 2c is that changelog should come at the very end of the index page, like release notes.

samccann:
Thanks for this! I like the current location of where you put the changelog. 

In the DaWGs meeting I mentioned that the left-hand nav is quite busy in the demo page. I'm not sure if this would be the same in the full docsite or not, but figured I'd mention it. @briantist mentioned it might benefit from using 'squash heirarchy'.

briantist:
> In the DaWGs meeting I mentioned that the left-hand nav is quite busy in the demo page. I'm not sure if this would be the same in the full docsite or not, but figured I'd mention it. briantist mentioned it might benefit from using 'squash heirarchy'.

Sorry my wording may have been confusing. The preview shown here, and its "busy" left-hand navigation, is _caused by_ using `--squash-hierarchy`, something we use sometimes for docs previews like this, where only a single collection is being rendered.

It won't look like that on the main docsite, or anywhere where the option is not explicitly used.

`--squash-hierarchy` is an existing option in `antsibull-docs` that isn't related to this changelog functionality.

samccann:
ah gotcha. thanks for clarifying!

felixfontein:
Now we have @oraNod who prefers the changelog at the bottom of the collection index page, and @samccann who likes the current location. @briantist what do you think? And are there other opinions on this? :)

samccann:
I posted on the docs channel as well in case anyone else has an opinion.

briantist:
I don't think I'm leaning too strongly in either direction. I have maybe a slight preference for its current location, but it doesn't matter that much. Mainly I wouldn't want it stuck somewhere in the middle, so near the top like it is now (just after communication), or all the way at the bottom, are decent options.

leogallego:
My take: 

As both the changelog and the plugins list can be quite long dependending on each collection:

As long as the changelog is not too long, and ideally if it only shows the "big changes/deprecation warning", I would prefer the middle, so you get a glimpe of changes just by scrolling through the doc before reaching the plugins. 

If it might be too long, then I would move it to the bottom, to avoid disrupting the plugin scrolling (which is most likely the reason you went to the docs in the first place)

felixfontein:
This is only about a *link* to the changelog. The changelog itself is on a separate page. So don't worry about its length :)

felixfontein:
In that particular example, the changelog is already in the 'middle'; it comes after the collection information (description, communication) and before the extra documentation (if available) and plugin/module/role index.

felixfontein:
Since most like the current location and there haven't been more opinions, I'm going to merge the current version. I'll wait some more days before creating a new release, so if compelling new opinions / insights come up we can still change it without breaking something that's released :-)

felixfontein:
Thanks everyone for your reviews and opinions!

PR #288, opened 2024-06-29
Collection links: allow to mention forums

felixfontein:
Right now we only mention Matrix, IRC, and mailing lists...


felixfontein:
(A next step would be updating the default list of communication channels for ansible-core - the Forum is added in the change in src/antsibull_docs/collection_links.py, but the others are still there. It might make sense to wait until https://forum.ansible.com/t/6812/7 is decided though. WDYT?)

oraNod:
> (A next step would be updating the default list of communication channels for ansible-core - the Forum is added in the change in src/antsibull_docs/collection_links.py, but the others are still there. It might make sense to wait until https://forum.ansible.com/t/6812/7 is decided though. WDYT?)

I feel like a good next step is to make additional changes that align with the recent communication guide overhaul from @Andersson007 Not quite sure how to approach it with antsibull-docs but this seems like a good first step in making sure the forum is the default in collection links. Also FYI @gundalow

oraNod (approved):
:rocket: Thank you @felixfontein

src/antsibull_docs/collection_links.py
```diff
@@ -70,6 +70,12 @@
                 "url": "https://groups.google.com/g/ansible-project",
             }
         ],
+        "forums": [
+            {
+                "topic": "Ansible Forum",
+                "url": "https://forum.ansible.com/c/help/6/none",
```
gundalow, line 76: Is this the default link that will show if `forums` isn't explicitly set?

felixfontein: No, this is only for the Communications sections of ansible.builtin (https://docs.ansible.com/ansible/devel/collections/ansible/builtin/#communication). Right now there are no default entries.

felixfontein:
Thanks for reviewing this! I'll work on a antsibull-docs release next, so this goes out and collections can start referencing the Forum (and potentially drop other communication channels in favor of it). If we want to add defaults (in certain cases), or adjust the list of communication channels for ansible.builtin / ansible-core, we can do that in follow-up releases.

PR #289, opened 2024-06-29
Improve error handling/rendering

felixfontein:
- Render errors as text code blocks, instead of using the default lexer (which happens to be YAML + Jinja2 for the Ansible docsite, for example, resulting in warnings).
- Improve handling of empty and broken changelogs.
- Remove leading spaces in RST paragraphs, to avoid them being interpreted as blockquotes.

### DIFF
--- a/src/antsibull_docs/jinja2/filters.py
+++ b/src/antsibull_docs/jinja2/filters.py
@@ -256,3 +256,33 @@ def rst_format(fmt: str, for_sphinx: bool = False) -> str:
 
 def column_width(string: str) -> int:
     return _column_width(string)
+
+
+def rst_indent(
+    value: str, width: t.Union[int, str], first: bool = False, blank: bool = False
+) -> str:
+    """Return a copy of the string with each line indented.
+
+    :param width: Number of spaces, or a string, to indent by.
+    :param first: Do not skip indenting the first line.
+    :param blank: Do not skip indenting empty lines.
+    """
+    if isinstance(width, str):
+        indent = width
+    else:
+        indent = " " * width
+
+    # The "\n" makes sure we handle an empty input or an input ending with a newline correctly
+    lines = (value + "\n").splitlines()
+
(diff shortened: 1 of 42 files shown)


felixfontein:
TODO:
```yaml
description: >
  Includes both the Generic Ip Pool and Tunnel Ip Pool.
  Generic - Used for general purpose within the network such as device
            management or communication between the network devices.
  Tunnel - Designated for the tunnel interfaces to encapsulate packets
           within the network protocol. It is used in VPN connections,
           GRE tunnels, or other types of overlay networks.
```
causes
```
WARNING: Block quote ends without a blank line; unexpected unindent.
```
Let's see whether we can improve this.

src/antsibull_docs/jinja2/filters.py
```diff
@@ -256,3 +256,33 @@ def rst_format(fmt: str, for_sphinx: bool = False) -> str:
...
+    if isinstance(width, str):
+        indent = width
+    else:
+        indent = " " * width
+
+    # The "\n" makes sure we handle an empty input or an input ending with a newline correctly
+    lines = (value + "\n").splitlines()
+
+    # Remove trailing whitespace
+    lines = [line.lstrip() for line in lines]
```
oraNod, line 279: Feels like a petty comment but could this have a more descriptive name like `stripped_lines` or something?

felixfontein: acbb7c89468aba5d733cfb68c78cf1c91fdf3a56

oraNod (approved):
LGTM. Thanks for making that one variable name more descriptive @felixfontein - I was also thinking about how to improve some of the readability here but that would probably mean refactoring the logic for handling the first line and blank lines. Anyway everything seems fine from what I can tell and I don't want to make a bunch of excessive comments. :smile_cat:

Thanks very much! :tada:

felixfontein:

PR #292, opened 2024-07-01
Collection docs linter: report empty markup

felixfontein:
Ref: #290

Empty markup (like `I()`, `L(,https://example.com)`, etc) makes no sense and usually indicates an error.


src/antsibull_docs/lint_plugin_docs.py
```diff
@@ -311,6 +311,82 @@ def _create_lookup(
...
+    def __init__(self, markup_validator: "_MarkupValidator", key: str):
+        self.markup_validator = markup_validator
+        self.errors = markup_validator.errors
+        self.key = key
+
+    def process_error(self, part: dom.ErrorPart) -> None:
+        self.errors.append(f"{self.key}: Markup error: {part.message}")
+
+    def process_bold(self, part: dom.BoldPart) -> None:
+        if not part.text:
```
oraNod, line 324: Would it be a good idea to call `strip()` for each of the checks for empty markup?

`if not part.text.strip():`

src/antsibull_docs/lint_plugin_docs.py
```diff
@@ -311,6 +311,82 @@ def _create_lookup(
     return lookup
 
 
+class _ValidationWalker(dom.NoopWalker):
+    def __init__(self, markup_validator: "_MarkupValidator", key: str):
+        self.markup_validator = markup_validator
+        self.errors = markup_validator.errors
+        self.key = key
+
```
oraNod, line 319: Maybe a helper method could be useful here to avoid some duplication? Something like:

```
def _append_error(self, source: str, message: str) -> None:
        self.errors.append(f"{self.key}: {source}: {message}")
```

src/antsibull_docs/lint_plugin_docs.py
felixfontein: Generally non-empty values that are only whitespace can be ok, especially for `C()` and `V()`. I only used `strip()` where whitespace only makes no sense, i.e. for URLs and RST references.

oraNod: makes sense. thanks for explaining!

src/antsibull_docs/lint_plugin_docs.py
felixfontein: Changed in 340cdb21a86d777563f2705163cd7838e12ba558.

oraNod: :raised_hands: I think I'm going to print this and frame it. I finally made a useful code-related suggestion to one of your PRs.

oraNod (approved):
Thanks @felixfontein

felixfontein:

PR #297, opened 2024-08-05
Remove pyre

felixfontein:
Let's get rid of it; mypy should be enough.


felixfontein:
CI will get fixed by #298.

PR #302, opened 2024-08-17
Make CLI startup faster, support tab completion with argcomplete

felixfontein:
Ref: https://github.com/ansible-community/antsibull-changelog/pull/173

### DIFF
--- a/changelogs/fragments/302-argcomplete.yml
+++ b/changelogs/fragments/302-argcomplete.yml
@@ -1,0 +1,5 @@
+minor_changes:
+  - "If you are using `argcomplete <https://pypi.org/project/argcomplete/>`__, you can now tab-complete ``antsibull-docs`` command lines.
+     See `Activating global completion <https://pypi.org/project/argcomplete/#activating-global-completion>`__ in the argcomplete README for
+     how to enable tab completion globally. This will also tab-complete Ansible commands such as ``ansible-playbook`` and ``ansible-test``
+     (https://github.com/ansible-community/antsibull-docs/pull/302)."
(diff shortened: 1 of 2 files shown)


src/antsibull_docs/cli/antsibull_docs.py
```diff
@@ -41,34 +52,33 @@
...
 # pylint: enable=wrong-import-position
 
 
 mlog = log.fields(mod=__name__)
 
+
+def _create_loader(module: str, function: str) -> Callable[[], Callable[[], int]]:
+    def load():
+        module_obj = import_module(f"antsibull_docs.cli.doc_commands.{module}")
+        return getattr(module_obj, function)
```
gotmax23, line 65: Does running `import_module` every time have much of a performance overhead?

changelogs/fragments/302-argcomplete.yml
```diff
@@ -0,0 +1,5 @@
+minor_changes:
+  - "If you are using `argcomplete <https://pypi.org/project/argcomplete/>`__, you can now tab-complete ``antsibull-docs`` command lines.
```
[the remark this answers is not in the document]
felixfontein: That doesn't apply since `argcomplete` wasn't there before.

src/antsibull_docs/cli/antsibull_docs.py
felixfontein: Line 64 is ran at most once (in total) per antsibull-docs invocation, so I don't think so.

gotmax23 (approved):

felixfontein:

Issue #306, opened 2024-08-20
Use of rsync inside generated build.sh prevents usage with readthedocs


ssbarnea:
To build with RTD, we will need to make use of custom build commands, aka `build.commands` feature, one that is highly promoted by them (aiming to become the default very soon).

Still, this feature comes with one major downside, losing ability to use `apt_packages` (and sudo is not available either).

All VM images provided by RTD do not have rsync installed and they [do not want to install it](https://github.com/readthedocs/readthedocs.org/issues/11219).

This means that in its current form antsibull-docs cannot be integration with RTD, which is currently the most common system that allows to test pull-requests changes made to documentation.

I do think that we might need to find a way to make rsync optional in `build.sh`, falling back to a plain old copy instead.

References:

- https://github.com/readthedocs/readthedocs.org/issues/11219
- https://github.com/readthedocs/readthedocs.org/issues/9599
- https://github.com/ansible/event-driven-ansible/pull/276

felixfontein:
This again is **not a bug**, but an enhancement.

ssbarnea:
@felixfontein We could argue four hours about what makes an issue a feature, bug or fug but that is better over a glass of beer. It becomes funny when a design flaw becomes a feature, just because the original design missed to cover that aspect. 🍻

felixfontein:
@ssbarnea it's pretty simple. If a feature works as designed, it's not a bug, but a feature / enhancement request. The feature here works as designed. So it's not a bug.

PR #307, opened 2024-08-20
Make use of rsync optional in build.sh

ssbarnea:
Fixes: #306

### DIFF
--- a/src/antsibull_docs/data/sphinx_init/build_sh.j2
+++ b/src/antsibull_docs/data/sphinx_init/build_sh.j2
@@ -4,11 +4,25 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 # Created with antsibull-docs @{ antsibull_docs_version }@
+# cspell: disable cprv
 
 set -e
 
 pushd "$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 trap "{ popd; }" EXIT
+
+# As rsync might be missing in places like RTD build system, we make it optional
+copy_dirs() {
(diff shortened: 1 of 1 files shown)


felixfontein (changes requested):
The whole reason of using rsync with the temp directory is that Sphinx only rebuilds the changed files in subsequent runs, and that files that should no longer be there are deleted. You cannot simply replace rsync by cp here without breaking both these things.

Instead of doing some silent fallback, I would implement this differently so that the user can pass a flag to `antsibull-docs sphinx-init` that directly lets antsibull-docs write to the destination instead of writing into a temp directory and using rsync. You can then use that option in CI when you start with a clean state anyway and don't need incremental builds.

felixfontein:
(To re-create the sphinx-init baseline, go to `tests/functional/` and run `./build-sphinx-init-baseline.sh`.)

src/antsibull_docs/data/sphinx_init/build_sh.j2
```diff
@@ -4,12 +4,26 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 # Created with antsibull-docs @{ antsibull_docs_version }@
+# cspell: disable cprv
```
gotmax23, line 7: We don't use cspell in this repository (whether or not we should is a separate discussion), so it'd be better not to have comments for it cluttering up the sources.

felixfontein:
I'm planning to do a new feature release of antsibull-docs this weekend. This would be a nice feature to include; when do you plan to continue working on it?

felixfontein:
#315 implements this differently.

PR #310, opened 2024-08-22
Adjust nox workflow to use local checkouts

felixfontein:
The workflow checks them all out, and then tells nox to install them directly from git... Tell nox to use the local checkouts. (That will also make it complain if we forgot to check out a project.)

### DIFF
--- a/.github/workflows/nox.yml
+++ b/.github/workflows/nox.yml
@@ -21,6 +21,7 @@ name: nox
 
 env:
   FORCE_COLOR: "1"
+  OTHER_ANTSIBULL_MODE: local
 
 jobs:
   nox:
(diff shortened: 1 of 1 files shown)


.github/workflows/nox.yml
```diff
@@ -77,10 +77,10 @@ jobs:
           python-versions: "${{ matrix.python-versions }}"
       - name: Set up nox environments
         run: |
-          OTHER_ANTSIBULL_MODE=git nox -v -e "${{ matrix.session }}" ${{ matrix.codecov && 'coverage' || '' }} --install-only
+          OTHER_ANTSIBULL_MODE=local nox -v -e "${{ matrix.session }}" ${{ matrix.codecov && 'coverage' || '' }} --install-only
       - name: "Run nox -e ${{ matrix.session }}"
         run: |
-          OTHER_ANTSIBULL_MODE=git nox -v -e "${{ matrix.session }}" --reuse-existing-virtualenvs --no-install
+          OTHER_ANTSIBULL_MODE=local nox -v -e "${{ matrix.session }}" --reuse-existing-virtualenvs --no-install
```
gotmax23, line 83: This works, of course, but what about adding `OTHER_ANTSIBULL_MODE=local` to `env` at the top instead? I think that might be cleaner.

felixfontein: I agree. Implemented in ac91c9ee0413b58c91dbfca5361badcbd2449403.

felixfontein:

Issue #141, opened 2023-05-03
Add deprecation marker to deprecated modules/plugins in indexes and plugin/module lists


felixfontein:

felixfontein:
And/or have a separate page listing everything that is deprecated.

felixfontein:
Where should the "deprecated content" list(s) live? Should there be one per collection, or even per plugin type per collection? Should there be a global "deprecated" list (for all collections)? Or one for every plugin type, but then for all collections?

oraNod:
I feel like a global "deprecated" list might be easier for community to find things and avoids fragmentation. Not sure though. This sounds like a good one for the next SC WG.

samccann:
Coming in late here - So back in the dim dawn of time, our 'all modules' pages (2.9 and earlier) used to mark each module with a (D) that was deprecated. We could consider doing that again (but with more dramatic deprecation marker. )

Or... we could ask on the forum! :-) 
I think the options are:

1. Add a deprecation marker within our collection/module indices pages
2. Create a "Deprecated plugins' page/list for all of them across all collections

I think we we do add markers to our index pages, it should be for all of them (main indices and per-collectoin index pages).

felixfontein:
@samccann 1. is already implemented by #320, see for example https://ansible.fontein.de/collections/index_module.html#community-docker (the removal versions won't be shown on docs.ansible.com to avoid confusion of Ansible versions with collection versions)

About 2.: either that, or having a list (one list per collection!) of deprecated plugins. Or split the list by plugin type (one file per plugin type). But yeah, a forum discussion sounds like a good idea. (For simplicity I'm currently tending to one page for all plugin types of all collections.)

PR #320, opened 2024-08-26
Show deprecation marker in module/plugin/role lists

felixfontein:
Fixes part of #141.

Looks like this:
image

I've decided not to show version numbers on the official docsite, since they easily confuse users (is this an Ansible version number? a collection version number?). WDYT?


oraNod (approved):
FWIW this looks good to me. I don't really see any code nits or anything and the color combos are fine from an [accessibility standpoint](https://webaim.org/resources/contrastchecker/). :shipit: 

Thanks @felixfontein

felixfontein:

PR #331, opened 2024-09-14
Avoid using pydantic compat layer where easily possible

felixfontein:
All other cases need v1 for easy linting (with extra checks).

### DIFF
--- a/src/antsibull_docs/schemas/app_context.py
+++ b/src/antsibull_docs/schemas/app_context.py
@@ -8,11 +8,10 @@
 import pydantic as p
 from antsibull_core.schemas.context import AppContext as CoreAppContext
 from antsibull_core.schemas.validators import convert_bool
-
-from antsibull_docs._pydantic_compat import Field
+from pydantic import Field
 
 #: Valid choices for the docs parsing backend
-DOC_PARSING_BACKEND_CHOICES_F = Field("auto", regex="^(auto|ansible-core-2\\.13)$")
+DOC_PARSING_BACKEND_CHOICES_F = Field("auto", pattern="^(auto|ansible-core-2\\.13)$")
 
 
 DEFAULT_COLLECTION_URL_TRANSFORM = (
--- a/src/sphinx_antsibull_ext/schemas/ansible_links.py
+++ b/src/sphinx_antsibull_ext/schemas/ansible_links.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 
 import typing as t
 
-from antsibull_docs._pydantic_compat import v1 as p
+import pydantic as p
 
 
 class AnsibleLink(p.BaseModel):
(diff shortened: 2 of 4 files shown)


src/antsibull_docs/schemas/app_context.py
```diff
@@ -8,11 +8,10 @@
 import pydantic as p
 from antsibull_core.schemas.context import AppContext as CoreAppContext
 from antsibull_core.schemas.validators import convert_bool
-
-from antsibull_docs._pydantic_compat import Field
+from pydantic import Field
 
 #: Valid choices for the docs parsing backend
-DOC_PARSING_BACKEND_CHOICES_F = Field("auto", regex="^(auto|ansible-core-2\\.13)$")
+DOC_PARSING_BACKEND_CHOICES_F = Field("auto", pattern="^(auto|ansible-core-2\\.13)$")
```
gotmax23, line 14: While the Rust regex create used by pydantic v2 to validate regexes is more efficient than Python's re that was used before, we should switch these to use `Literal` instead, either here in another PR.

gotmax23 (commented):
This looks good after a cursory look, but we should consider getting rid of the unnecessary regex usage for performance/simplicity reasons.

src/sphinx_antsibull_ext/schemas/ansible_links.py
```diff
@@ -18,13 +18,14 @@ class AnsibleLink(p.BaseModel):
...
 
-    @p.root_validator()
-    # pylint:disable=no-self-argument
-    def one_of_url_and_ref(cls, values: dict[str, t.Any]) -> dict[str, t.Any]:
-        has_url = values.get("url")
-        has_ref = values.get("ref")
-        if has_url == has_ref:
-            raise ValueError("Exactly one of 'url' and 'ref' must be specified.")
+    @p.model_validator(mode="before")
+    @classmethod
```
gotmax23, line 22: I don't think classmethod is strictly necessary from pydantic's side. Do one of our linters/type checkers complain if it's not there?

src/antsibull_docs/schemas/app_context.py
felixfontein: I agree. But let's do that in another PR.

src/sphinx_antsibull_ext/schemas/ansible_links.py
felixfontein: I don't know whether it's really necessary (I haven't tried), but the (only) example in the pydantic documentation also uses it: https://docs.pydantic.dev/latest/concepts/validators/#model-validators

felixfontein: (Also it avoids problems with other linters who don't know that pydantic does this.)

gotmax23: > @field_validators are "class methods", so the first argument value they receive is the UserModel class, not an instance of UserModel. We recommend you use the @classmethod decorator on them below the @field_validator decorator to get proper type checking.

I suppose the same thing applies here.

Marking as resolved.

gotmax23 (approved):

felixfontein:

gotmax23:
Haha, was about to push that button

gotmax23:
Thanks!

src/antsibull_docs/schemas/app_context.py
felixfontein: Hmm, I tried replacing `doc_parsing_backend: str = DOC_PARSING_BACKEND_CHOICES_F` with `doc_parsing_backend: str = t.Literal["auto", "ansible-core-2.13"]`, and fed it a config file with an invalid value. With `t.Literal` it did **not** complain about the invalid value.

felixfontein: Hmm, and I tried it again with the CollectionMetadata schema from antsibull-core, and *there* it works... I have to try again :)

felixfontein: Ok, I did a stupid mistake. It works fine...

PR #333, opened 2024-09-15
Migrate remaining schemas to pydantic 2

felixfontein:
Currently contains #331. Will rebase once that has been merged.

- [x] Rebase to remove #331 once merged.
- [x] Remove `TEMP` commit once https://github.com/ansible-community/antsibull-core/pull/169 has been merged.
- [x] Release new antsibull-core version with https://github.com/ansible-community/antsibull-core/pull/169.
- [x] Bump antsibull-core dependency and remove `TEMP` commit once antsibull-core 3.2.0 is out.
- [x] Migrate docs schemas to pydantic 2.


felixfontein:
@gotmax23 I managed to convert all pydantic v1 models to pydantic v2. That also means that https://github.com/ansible-community/antsibull-core/pull/169 is sufficient as-is and does not need more changes for the documentation schemas.

src/antsibull_docs/process_docs.py
```diff
@@ -40,6 +41,48 @@ def get_collection_namespaces(collection_names: Iterable[str]) -> dict[str, list
...
+def _exc_to_string(exc: p.ValidationError, model_name: str) -> str:
+    def _display_error_loc(error: pydantic_core.ErrorDetails) -> str:
+        # We sort out all strings starting with an upper-case letter
+        # since pydantic sometimes inserts class names, for example:
+        #
+        #   entry_points -> main -> seealso -> 1 -> SeeAlsoPluginSchema -> extra
+        #
+        # Since our attribute names are all lower-case and our class
+        # names all start with an upper-case letter, this effectively
+        # removes all class names from the location sequence.
```
felixfontein, line 56: Hmm, the reasoning doesn't work due to how `options` etc. work - the keys there are arbitrary strings that users provide. There exist modules whose option names are starting with upper-case letters :(

For example, amazon.aws.elb_application_lb has a (nested) option `listeners.DefaultActions[].TargetGroupName`.

felixfontein: Fixed in adcc19fdda1feaad869a2ac594318fad21c6f8e0.

src/antsibull_docs/collection_config.py
```diff
@@ -50,10 +50,10 @@ async def load_collection_config(
         config_path = os.path.join(collection_path, "docs", "docsite", "config.yml")
         if os.path.isfile(config_path):
             try:
-                return CollectionConfig.parse_obj(load_yaml_file(config_path))
-            except v1.ValidationError:
+                return CollectionConfig.model_validate(load_yaml_file(config_path))
+            except p.ValidationError:
                 pass
-        return CollectionConfig.parse_obj({})
+        return CollectionConfig.model_validate({})
```
gotmax23, line 56:
```suggestion
        return CollectionConfig()
```

should be equivalent, but more efficient

src/antsibull_docs/collection_links.py
```diff
@@ -130,9 +130,9 @@ def load(
     else:
         ld = {}
     try:
-        result = CollectionLinks.parse_obj(ld)
-    except v1.ValidationError:
-        result = CollectionLinks.parse_obj({})
+        result = CollectionLinks.model_validate(ld)
+    except p.ValidationError:
+        result = CollectionLinks.model_validate({})
```
gotmax23, line 135:
```suggestion
        result = CollectionLinks()
```

src/antsibull_docs/docs_parsing/__init__.py
```diff
@@ -114,4 +114,6 @@ def __repr__(self):
 
     @classmethod
     def empty(cls, path="."):
-        return cls(path=path, docs_config=CollectionConfig.parse_obj({}), version=None)
+        return cls(
+            path=path, docs_config=CollectionConfig.model_validate({}), version=None
```
gotmax23, line 118:
```suggestion
            path=path, docs_config=CollectionConfig(), version=None
```

src/antsibull_docs/process_docs.py
```diff
@@ -40,6 +43,78 @@ def get_collection_namespaces(collection_names: Iterable[str]) -> dict[str, list
     return namespaces
 
 
+_CAPITAL_LETTERS = tuple("ABCDEFGHIJKLMNOPQRSTUVWXYZ")
```
gotmax23, line 46: See `string.ascii_uppercase` from the stdlib

src/antsibull_docs/process_docs.py
```diff
@@ -63,36 +138,45 @@ def normalize_plugin_info(
 
     errors: list[str] = []
     if plugin_type == "role":
+        role_schema: type[BaseModel] = DOCS_SCHEMAS[
+            plugin_type
+        ]  # type: ignore[attr-defined, assignment]
```
gotmax23, line 143: Why does this need a `type: ignore`? Is `DOC_SCHEMAS` not typed correctly?

src/antsibull_docs/process_docs.py
```diff
@@ -63,36 +138,45 @@ def normalize_plugin_info(
...
-            parsed = DOCS_SCHEMAS[plugin_type].parse_obj(plugin_info)  # type: ignore[attr-defined]
-            return parsed.dict(by_alias=True), errors
-        except v1.ValidationError as e:
-            raise ValueError(str(e))  # pylint:disable=raise-missing-from
+            parsed = role_schema.model_validate(plugin_info)
+            return parsed.model_dump(by_alias=True), errors
+        except p.ValidationError as e:
+            raise ValueError(  # pylint:disable=raise-missing-from
+                _exc_to_string(e, role_schema.__name__)
+            )
```
gotmax23, line 150: Why doesn't this have a `from`?

src/antsibull_docs/process_docs.py
```diff
@@ -63,36 +138,45 @@ def normalize_plugin_info(
...
             # Note: We looped through doc first and returned an exception if doc did not normalize
             # so we're able to use it in the error message here.
             errors.append(
                 f'Unable to normalize {new_info["doc"]["name"]}: {field}'
-                f" due to: {str(e)}"
+                f" due to: {_exc_to_string(e, schema.__name__)}"
             )
 
-            field_model = DOCS_SCHEMAS[plugin_type][field].parse_obj({})  # type: ignore[index]
+            field_model = DOCS_SCHEMAS[plugin_type][field].model_validate({})  # type: ignore[index]
```
gotmax23, line 177:
```suggestion
            field_model = DOCS_SCHEMAS[plugin_type][field]()  # type: ignore[index]
```

src/antsibull_docs/schemas/docs/ansible_doc.py
```diff
@@ -78,64 +81,54 @@ class AnsibleDocSchema(BaseModel):
...
-        a dynamic key which we can't automatically map to an attribute name.
-    """
+GenericPluginSchema = p.RootModel[dict[str, PluginSchema]]
+"""
+Document the output of ``ansible-doc -t PLUGIN_TYPE PLUGIN_NAME``.
 
-    __root__: dict[str, CallbackSchema]
+.. note:: Both the model and the dict will be wrapped in an outer dict with your data mapped
+    to the ``__root__`` key. This happens because the toplevel key of ansible-doc's output is
+    a dynamic key which we can't automatically map to an attribute name.
```
gotmax23, line 90: Is this still the case with the way new `RootModel`s work?

src/antsibull_docs/schemas/docs/plugin.py
```diff
@@ -208,16 +206,17 @@ class PluginExamplesSchema(BaseModel):
...
     examples_format: str = "yaml"
 
-    @p.validator("examples_format", always=True)
-    # pylint:disable=no-self-argument
-    def extract_examples_format(cls, value: t.Any, values: dict[str, t.Any]):
-        if isinstance(examples := values.get("examples"), str):
-            if fmt_match := _EXAMPLES_FMT_RE.match(examples.lstrip()):
-                value = fmt_match.group(1)
-        return value
+    @p.model_validator(mode="before")
```
gotmax23, line 209: Why not use a `field_validator`? That the equivalent of the old `validator` API.

src/antsibull_docs/process_docs.py
felixfontein: `plugin_type` is not typed correctly. It's an arbitrary string at this point.

src/antsibull_docs/process_docs.py
felixfontein: Because we didn't have a `from` in the existing code as well.

src/antsibull_docs/process_docs.py
felixfontein: I don't think this is a good idea as long as we use `Field()` with a default value.

src/antsibull_docs/schemas/docs/plugin.py
felixfontein: I wasn't able to get hold of the other field's values in a `field_validator`.

Also conceptually field validator doesn't really make sense since this sets another field based on a field's value, which is not what a field validator should be doing.

src/antsibull_docs/process_docs.py
felixfontein: f3f1f25360de562bb5f027ed339a70b3021ca505

src/antsibull_docs/schemas/docs/ansible_doc.py
felixfontein: No, I forgot to update the comments. I removed the notes in ff7148607b96bcc0046e8f308d0d09e963c58ae0.

src/antsibull_docs/process_docs.py
gotmax23: How does `model_validate()` affect Fields with default values?

felixfontein: It should assign the default value from the `Field` instance to them if the field isn't provided the input.

src/antsibull_docs/process_docs.py
gotmax23: I guess we can leave it as is and fix it later, but python does generate a confusing chain when you raise an Exception in an Exception handler without a `from`.

src/antsibull_docs/process_docs.py
gotmax23: It should do the same thing if you call the constructor without parameters, no?

src/antsibull_docs/schemas/docs/plugin.py
gotmax23: :+1:

gotmax23:
Looks like everything except coverage (with a less than 1% difference...) is passing. I'll do one last look through, and also try to run a test package docs build in ansible-documentation's CI.

src/antsibull_docs/process_docs.py
felixfontein: I would hope so, but I don't know. After my previous comment I wanted to try it out, and noticed that I can't because all structs that use such fields are not default-constructable (due to some validators). Which also means that all structs where the above construct is used do not use `Field` with a `default` value, so it doesn't really matter and your replacement is fine :)

src/antsibull_docs/process_docs.py
felixfontein: True. The exceptions here never surface to the user though, since they are caught and their string representation stored.

(Which also means that adding `from` isn't a problem.)

gotmax23 (approved):
You've addressed my feedback, and I looked at this pretty closely the first time, so ACK, pending the new antsibull-core release and version pin updates.

felixfontein:

PR #338, opened 2024-09-20
CI: Replace sensu.sensu_go with community.sops

felixfontein:
Back then I chose sensu.sensu_go as the only collection included in Ansible that has roles. It's no longer the only one (https://docs.ansible.com/ansible/devel/collections/index_role.html), and it doesn't seem to be actively maintained (Ref: https://forum.ansible.com/t/8380), so I want to replace it with another collection.


felixfontein:
#339 will fix the failing lint check.

PR #405, opened 2025-07-18
Remove ansible._protomatter if it wasn't explicitly requested and has no content

felixfontein:
Right now it shows up on the collection list for the official docsite, but has no content.
(The content is actively hidden by ansible-doc: https://github.com/ansible/ansible/blob/38823665858c57ffd1fbc08c6acbead20d339c51/lib/ansible/cli/doc.py#L799-L801)


oraNod:
Thanks @felixfontein

